PR #60, opened 2020-09-25
Set initial coin supply

zeptin:
This needs to be applied to the test & regtest networks as well before launch


zeptin:
Reopening as #64 due to a merge conflict

PR #94, opened 2020-10-15
Fix claimer logic

zeptin:


src/Stratis.Features.FederatedPeg/Distribution/RewardClaimer.cs
```diff
@@ -63,11 +63,11 @@ private void OnBlockConnected(BlockConnected blockConnected)
 
             // As this runs on the mainchain we presume there will be a coinstake transaction in the block (but during the PoW era there obviously may not be).
             // If not, just do nothing with this block.
-            if (maturedBlock.Transactions.Count < 2 || !blockConnected.ConnectedBlock.Block.Transactions[1].IsCoinStake)
+            if (maturedBlock.Transactions.Count < 2 || !maturedBlock.Transactions[1].IsCoinStake)
```
quantumagi, line 66: Is it possible that `maturedBlock` can be `null` ?

fassadlr: @quantumagi it looks like we do a check above to see if the tip is less than the `minStakeConfirmations`. This will prevent a null ChainedHeader.

fassadlr (approved):

quantumagi:
image

quantumagi (changes requested):
I got an error on the mainchain while syncing the federation wallet from scratch.

zeptin:
LGTM, I will address the fee issue in another PR. I think perhaps adding a bit of logic to the mempool will be the easiest.

quantumagi (approved):

PR #119, opened 2020-10-26
Various Reward Distribution Fixes

fassadlr:


src/Stratis.Features.FederatedPeg/TargetChain/WithdrawalTransactionBuilder.cs
```diff
@@ -70,18 +69,19 @@ public Transaction BuildWithdrawalTransaction(int blockHeight, uint256 depositId
                     IgnoreVerify = true,
                     WalletPassword = walletPassword,
                     Sign = sign,
-                    Time = this.network.Consensus.IsProofOfStake ? blockTime : (uint?) null
+                    Time = this.network.Consensus.IsProofOfStake ? blockTime : (uint?)null
                 };
 
-                if (!this.federatedPegSettings.IsMainChain && recipient.ScriptPubKey == StraxCoinstakeRule.CirrusTransactionTag)
+                // Withdrawals from the sidechain won't have the OP_RETURN transaction tag, so we need to check against the ScriptPubKey of the Cirrus Dummy address.
```
zeptin, line 75: I think we should also document what the transaction tag actually is versus the dummy address to prevent future confusion. The transaction tag appears on the mainchain to indicate that the transaction is initiating a cross-chain transfer of the reward. Then when the deposit is extracted the withdrawal has the dummy address as the recipient (as this PR correctly fixes) as a placeholder value to indicate that the actual recipient list needs to be obtained from the distribution manager.

zeptin (approved):

zeptin (approved):

PR #121, opened 2020-10-27
Improvements to vote weight determination

zeptin:
This leverages the ability to look up the last balance decrease transaction when computing vote weight. If the decrease was due to a burn transaction the burn amount (if greater) will be used instead of the original vote weight.

### DIFF
--- a/src/SwapExtractionTool/VoteExtractionService.cs
+++ b/src/SwapExtractionTool/VoteExtractionService.cs
@@ -7,6 +7,7 @@ using Flurl;
 using Flurl.Http;
 using NBitcoin;
 using Stratis.Bitcoin.Controllers.Models;
+using Stratis.Bitcoin.Features.BlockStore.Controllers;
 using Stratis.Bitcoin.Features.BlockStore.Models;
 
 namespace SwapExtractionTool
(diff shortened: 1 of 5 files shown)


src/SwapExtractionTool/VoteExtractionService.cs
```diff
@@ -131,11 +136,33 @@ private async Task ProcessBlockForCollateralVoteTransactionsAsync(BlockTransacti
...
+                    LastBalanceDecreaseTransactionModel lastBalanceDecreaseTransaction = await $"http://localhost:{this.StratisNetworkApiPort}/api"
+                        .AppendPathSegment($"blockstore/{BlockStoreRouteEndPoint.GetLastBalanceDecreaseTransaction}")
+                        .SetQueryParams(new { addresses = potentialStratAddress, minConfirmations = 0 })
+                        .GetJsonAsync<LastBalanceDecreaseTransactionModel>();
+
+                    if (lastBalanceDecreaseTransaction.BlockHeight > blockHeight)
+                    {
+                        foreach (Vout txOut in lastBalanceDecreaseTransaction.Transaction.VOut)
+                        {
+                            if (txOut.ScriptPubKey.Type == "nulldata" && Money.Coins(txOut.Value) > determinedBalance)
```
fassadlr, line 156: I'm wondering whether we should also check the burn transaction itself to see if it does contain a STRAX address? Maybe a bit overkill? 🤔

fassadlr (approved):

PR #127, opened 2020-10-28
Update vote extraction

zeptin:
Instead of trying to maintain all the necessary indexes locally, this leverages the block explorer for looking up the inputs of burn transactions. If it recognises that an address with an existing vote funded the burn transaction, it amends the vote weight for that address accordingly. If multiple vote addresses are in the inputs of a burn, only 1 will be assigned the total value of the burn and the rest are reset to their final detected balance per the address indexer.

### DIFF
--- a/src/SwapExtractionTool/VoteExtractionService.cs
+++ b/src/SwapExtractionTool/VoteExtractionService.cs
@@ -5,7 +5,6 @@ using System.Text;
 using System.Threading.Tasks;
 using Flurl;
 using Flurl.Http;
-using Microsoft.AspNetCore.Server.IIS.Core;
 using NBitcoin;
 using Stratis.Bitcoin.Controllers.Models;
 using Stratis.Bitcoin.Features.BlockStore.Controllers;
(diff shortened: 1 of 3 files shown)


src/SwapExtractionTool/VoteExtractionService.cs
```diff
@@ -5,6 +5,7 @@
 using System.Threading.Tasks;
 using Flurl;
 using Flurl.Http;
+using Microsoft.AspNetCore.Server.IIS.Core;
```
zeptin, line 8:
```suggestion
```

Bump CI

src/SwapExtractionTool/VoteExtractionService.cs
```diff
@@ -103,6 +99,62 @@ private async Task ProcessBlockForCollateralVoteTransactionsAsync(BlockTransacti
...
+                    // Check the inputs of the transaction to see if it was funded by one of the vote addresses.
+                    string address = null;
+                    foreach (In input in tx._in)
+                    {
+                        if (input.hash == null)
+                            continue;
+
+                        // The block explorer calls this 'hash' but it is in fact the address that funded the input.
+                        // It is possible that several voting addresses consolidated together in one swap transaction, so just take the first one.
+                        if (this.collateralVotes.ContainsKey(input.hash))
```
fassadlr, line 119: Kevin, I cant see how this will ever be true as we only ever add to/change the `collateralVotes` array if `address` is not null but then we only set `address` to something if the array contains the `input.hash`?

fassadlr (approved):

PR #150, opened 2020-11-07
Allow external API access to docker container

zeptin:


zeptin:
Whoops, mislabeled this

zeptin:
Probably best to regard this as WIP for now, it works locally but there are inconsistent results

zeptin:
I have rechecked this and it is definitely not possible to access the docker container's node API from another machine without this change.

With the conf changed the API is accessible from both localhost and remotely.

Note that the conf file is not actually copied to the container by default; but the readme contains instructions for how to do this.

PR #165, opened 2020-11-11
Enable kicking should be an command line option

fassadlr:


src/Stratis.Features.FederatedPeg/TargetChain/CrossChainTransferStore.cs
```diff
@@ -665,13 +663,13 @@ public Task<Transaction> MergeTransactionSignaturesAsync(uint256 depositId, Tran
 
                                 if (this.ValidateTransaction(transfer.PartialTransaction, true))
                                 {
-                                    this.logger.LogInformation("Deposit: {0} collected enough signatures and is FullySigned", transfer.DepositTransactionId);
+                                    this.logger.LogDebug("Deposit: {0} collected enough signatures and is FullySigned", transfer.DepositTransactionId);
```
fassadlr, line 666: These should be `debug` logs.

src/Stratis.Features.FederatedPeg/TargetChain/CrossChainTransferStore.cs
```diff
@@ -602,8 +602,6 @@ public Task<Transaction> MergeTransactionSignaturesAsync(uint256 depositId, Tran
                         if (!this.Synchronize())
                             return null;
 
-                        this.logger.LogInformation("Merging signatures for deposit : {0}", depositId);
```
fassadlr, line 605: Was used during testing, should be removed.

quantumagi:
You should probably disable the join federation API as well.
Also, what is your opinion on this?:
```
                        // Member that were about to be kicked when voting started don't participate.
                        if (this.idleFederationMembersKicker != null)
                        {
                            ChainedHeader chainedHeader = chBlock.ChainedHeader.GetAncestor(poll.PollStartBlockData.Height);
                            foreach (string pubKey in fedMembersHex)
                            {
                                if (this.idleFederationMembersKicker.ShouldBeKicked(new PubKey(pubKey), chainedHeader.Header.Time, out _))
                                {
                                    fedMembersHex.TryRemove(pubKey);
                                }
                            }
                        }
```
and `int requiredVotesCount = (fedMembersHex.Count / 2) + 1;`

As a result of the larger quorum requirement this change will disable voting altogether.
An alternative way to approach this would have been with a new flag: DisableDynamicMembership that laser-targets the code to disable:
(1) `if (!DisableDynamicMembership && this.ShouldBeKicked(fedMemberToActiveTime.Key, timestamp, out uint inactiveForSeconds))`
(2):
```
        public async Task<IActionResult> JoinFederationAsync([FromBody] JoinFederationRequestModel request, CancellationToken cancellationToken = default)
        {
            Guard.NotNull(request, nameof(request));
            // Checks that the request is valid.
            if (!this.ModelState.IsValid)
            {
                this.logger.LogTrace("(-)[MODEL_STATE_INVALID]");
                return ModelStateErrors.BuildErrorResponse(this.ModelState);
            }
            try
            {
```
Add `if (DisableDynamicMembership) throw...`

However if we don't need voting for whitelisting smart contracts then it probably does not matter (apart from being a slightly more risky change).

quantumagi:
After this change the quorum requirement will increase from 4 to 8. Please confirm that this is not a problem, i.e. there is no requirement to white-list smart contract hashes during this period, then I can approve this PR.

quantumagi (approved):

quantumagi:
image

PR #179, opened 2020-11-15
Suppress NULL fail errors

fassadlr:
These are only indicating that the multisig transaction is not fully signed yet, so it is a bit misleading.


quantumagi (approved):
LGTM

PR #200, opened 2020-11-20
Implement fundrawtransaction and signrawtransaction RPC

zeptin:
There are two issues to be resolved:

1. The `ToHex()` helper method in the RPC client needs to return valid hex for transactions with empty input lists. It is currently hardcoded, but this may fail with proper segwit transactions.
2. The RPC middleware does not seem to be propagating the fund raw transaction options model to the controller. So one of the tests is failing due to the change position not being properly specified.

There are also several unimplemented option settings for `fundrawtransaction`, such as the fee rate, but the initial implementation works.


zeptin:
This includes the following:

1. `fundrawtransaction` and `signrawtransaction` RPC methods, with basic options supported for FRT.
2. Offline transaction signing API endpoints that allow a synced wallet restored via extpubkey (i.e. with no private keys) to create a template transaction for an offline/air-gapped node to sign with its private keys.
3. `newaddresses` API endpoint for the use case where a new address needs to be returned with every call rather than returning the first available unused address.

src/Stratis.Bitcoin.Features.RPC/FundRawTransactionResponse.cs
```diff
@@ -1,17 +1,23 @@
 ﻿using NBitcoin;
+using Newtonsoft.Json;
 
 namespace Stratis.Bitcoin.Features.RPC
 {
     public class FundRawTransactionResponse
     {
+        [JsonProperty(PropertyName = "hex")]
```
fassadlr, line 8: Should this be `hex` ?

src/Stratis.Bitcoin.Features.RPC/SignRawTransactionResponse.cs
```diff
@@ -0,0 +1,22 @@
+﻿using NBitcoin;
+using Newtonsoft.Json;
+
+namespace Stratis.Bitcoin.Features.RPC
+{
+    public class SignRawTransactionResponse
+    {
+        [JsonProperty(PropertyName = "hex")]
```
fassadlr, line 8: Same here, should this be `hex` ?

src/Stratis.Bitcoin.Features.Wallet/Models/BuildOfflineSignResponse.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// The account containing the UTXO(s) and address(es) needed to sign the transaction offline.
+        /// </summary>
+        [JsonProperty(PropertyName = "walletAccount")]
+        public string WalletAccount { get; set; }
+
+        /// <summary>
+        /// The transaction that needs to be signed by the offline node.
+        /// </summary>
+        [JsonProperty(PropertyName = "unsignedTransaction")]
+        public string UnsignedTransaction { get; set; }
```
fassadlr, line 27: I assume this will be a hex?

src/Stratis.Bitcoin.Features.Wallet/Services/WalletService.cs
```diff
@@ -1196,6 +1196,192 @@ public async Task<List<string>> Sweep(SweepRequest request, CancellationToken ca
...
+                    recipients.Add(new Recipient
+                    {
+                        ScriptPubKey = destination,
+                        Amount = recipientModel.Amount
+                    });
+                }
+
+                // If specified, get the change address, which must already exist in the wallet.
+                HdAddress changeAddress = null;
+                if (!string.IsNullOrWhiteSpace(request.ChangeAddress))
```
fassadlr, line 1239: If we have specified a change address, shouldn't this also throw? Or can a null change address be specified?

src/Stratis.Bitcoin.Features.Wallet/Services/WalletService.cs
```diff
@@ -1196,6 +1196,192 @@ public async Task<List<string>> Sweep(SweepRequest request, CancellationToken ca
...
+                    Addresses = addresses
+                };
+            }, cancellationToken);
+        }
+
+        public async Task<WalletBuildTransactionModel> OfflineSignRequest(OfflineSignRequest request, CancellationToken cancellationToken)
+        {
+            return await Task.Run(() =>
+            {
+                Transaction unsignedTransaction = this.network.CreateTransaction(request.UnsignedTransaction);
```
fassadlr, line 1330: Ah yes OK it is a hex, so maybe rename to `UnsignedTransactionHex`?

fassadlr (approved):
Looks good @zeptin and thanks for adding tests. Just a few questions but if they don't require any attention we can merge.

src/Stratis.Bitcoin.Features.RPC/FundRawTransactionResponse.cs
zeptin: Yes; the actual field returned by bitcoind is named `hex`. I didn't want to have to make unnecessary changes to NBitcoin, so the name stayed.

src/Stratis.Bitcoin.Features.RPC/SignRawTransactionResponse.cs
zeptin: Similarly for this one

src/Stratis.Bitcoin.Features.Wallet/Services/WalletService.cs
zeptin: Yes it can be null, in that case the transaction build context will find one from the wallet.

PR #258, opened 2020-12-07
Offline cold staking setup

zeptin:
This extends the offline transaction signing capability of the node to cold staking setup transactions.

The prerequisites are:
1. The offline node, with the full cold wallet restored (does not need to be synced)
2a. The online node, with the wallet from (1) restored via extpubkey (i.e. specifically the cold address account, wallet needs to be synced)
2b. The online node, with the full hot wallet restored (needs to be synced)

The hot wallet needs to be from a separate mnemonic to the offline cold wallet, as it is extremely messy to add a hot account onto the cold extpubkey wallet. Additionally, in order to actually be able to cold stake the hot address private key needs to be available to the online node.

An end-to-end integration test is included demonstrating the rough sequence that is followed in order to be able to complete an offline cold staking setup.

### DIFF
--- a/src/NBitcoin/Script.cs
+++ b/src/NBitcoin/Script.cs
@@ -344,7 +344,8 @@ namespace NBitcoin
         P2WPKH,
         P2WSH,
         MultiSig,
-        Federation
+        Federation,
+        ColdStaking
     }
 
     public class ScriptSigs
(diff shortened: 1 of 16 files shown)


fassadlr (approved):

fassadlr:
@zeptin PR looks good, testing will show any issues :)

PR #288, opened 2020-12-14
Simplify offline cold staking setup

zeptin:
It is now no longer required that the 'cold address' account be loaded on the online node via extpubkey. Instead the cold staking scriptPubKey is computed directly from the supplied hot and cold addresses, whether they are present in the node's wallet or not.

### DIFF
--- a/src/Stratis.Bitcoin.Features.ColdStaking/ColdStakingManager.cs
+++ b/src/Stratis.Bitcoin.Features.ColdStaking/ColdStakingManager.cs
@@ -370,7 +370,7 @@ namespace Stratis.Bitcoin.Features.ColdStaking
             }
             else
             {
-                // In offline mode we relax all the restrictions to enable simpler setup.
+                // In offline mode we relax all the restrictions to enable simpler setup. The user should ensure they are using separate wallets, or the cold private key could be inadvertently loaded on the online node.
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin.Features.ColdStaking/ColdStakingManager.cs
```diff
@@ -326,56 +326,71 @@ internal HdAddress GetFirstUnusedColdStakingAddress(string walletName, bool isCo
...
+                {
+                    hotPubKeyHash = new BitcoinPubKeyAddress(hotWalletAddress, wallet.Network).Hash;
+                    coldPubKeyHash = new BitcoinPubKeyAddress(coldWalletAddress, wallet.Network).Hash;
+                }
             }
             else
             {
-                coldAccount = this.GetOrCreateColdStakingAccount(walletName, true, walletPassword, extPubKey);
-                hotAccount = this.GetOrCreateColdStakingAccount(walletName, false, walletPassword, extPubKey);
+                // In offline mode we relax all the restrictions to enable simpler setup.
```
zeptin, line 373:
```suggestion
                // In offline mode we relax all the restrictions to enable simpler setup. The user should ensure they are using separate wallets, or the cold private key could be inadvertently loaded on the online node.
```
Boot CI

fassadlr (approved):

PR #327, opened 2020-12-29
Change full node console window title to show which full node is running

0tim0:
Added a line to the node start up to set the console window title to identify which full node this is (for clarity when running Strax and Cirrus on the same machine, especially thinking of MNs)

(Note: I've verified this works on my own machine, but full disclosure that I'm new to C# and most of my education has been reverse engineering code in this repo)

### DIFF
--- a/src/Stratis.CirrusD/Program.cs
+++ b/src/Stratis.CirrusD/Program.cs
@@ -35,6 +35,8 @@ namespace Stratis.CirrusD
         {
             try
             {
+                // set the console window title to identify this as a Cirrus full node (for clarity when running Strax and Cirrus on the same machine)
+				Console.Title = "Cirrus full node";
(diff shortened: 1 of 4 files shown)


src/Stratis.CirrusD/Program.cs
```diff
@@ -35,6 +35,8 @@ public static async Task MainAsync(string[] args)
         {
             try
             {
+                // set the console window title to identify this as a Cirrus full node (for clarity when running Strax and Cirrus on the same machine)
+                Console.Title = "Cirrus full node";
```
fassadlr, line 39: Hi @0tim0, if you want this for the masternode's then should also be set in CirrusMinerD and CirrusPegD

fassadlr: Also, please capitalize `full node`. Same with Strax below.

0tim0: @fassadlr Thanks for the feedback.  Updates made as requested

fassadlr (approved):

PR #328, opened 2020-12-30
Make sweep work for less than 500 outputs

sondreb:
- This fixes an issue where the sweep would only work if there was more than 500 outputs.
- The "currentOutputCount" is not really used and is reset for every single UTXOs discovered and individual transactions is generated for each UTXOs, they are not combined into a single transaction.
- The sweep might need some refactoring, and the FallbackFee might need to be increased a little bit to work properly and not generate transactions with too low fee during sweep.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Services/WalletService.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Services/WalletService.cs
@@ -1148,7 +1148,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Services
                     // Not many wallets will have this many inputs, but we have to ensure that the resulting transactions are
                     // small enough to be broadcast without standardness problems.
                     // Since there is only 1 output the size of the inputs is the only consideration.
-                    if (total == 0 || currentOutputCount < 500)
+                    if (total == 0 || currentOutputCount > 500)
                         continue;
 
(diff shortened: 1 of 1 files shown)


sondreb:
I think the original thinking behind this sweep code was to grab all UTXOs and generate a single transaction, but it just doesn't do that right now so this is a tiny fix to make it actually work.

zeptin:
No, the purpose of the `<` was to keep adding UTXOs to the builder until the transaction has that many inputs, as a rough heuristic for the maximum possible number. A more robust but expensive approach is to estimate the transaction size after each addition, as different types of transactions have differing scriptSig sizes.

Once a large enough transaction is reached the remainder of the loop executes. You thus get a list of transactions with 500 inputs at the end. So I think more of a refactor would be needed in order to handle the case where there aren't at least 500 inputs. Probably the latter part of the loop should be moved to a method, and it should be executed once more once the loop terminates to make a final transaction with the remaining UTXOs.

sondreb:
So this small PR will fix it so that it accepts a sweep, but there is more to be done in the implementation I think. I might do some more improvements and make another PR, unless someone else does it. Additionally I had to increase the FallbackFee a little bit to ensure it used enough fee to be approved by the policy. Fee should probably be added as an parameter to the API.

sondreb:
Made an update to the sweep, please review =)

PR #334, opened 2021-01-04
Improve Federation Wallet and CCTS sync speed

fassadlr:
In this PR there are 2 major improvements:

1. The reward distribution builder walks back the chain to determine the encoded commitment heights for each block and does so for each deposit block received. This means that it unnecessarily walks over the same part of the chain multiple times. We can cache the set so that the lookup is faster. This improved the build withdrawal reward transaction code from 11000ms to 400ms.
2. The federation wallet gets saved unnecessarily often. On some code paths it gets saved 3-4 in a row. This is OK if the wallet remains small but once the reward transactions start coming in grows quickly and a save can task 100-150ms. I've changed this so that caller checks if any wallet modifications were made before saving.


zeptin (approved):
Nice

fassadlr:
I synced up CirrusTest's federation wallet in 2h30min.

quantumagi (approved):
LGTM

PR #356, opened 2021-01-11
postman requests for test script error fixed

lisa3907:
Fixed some scripts that are not the same as the actual API call method. 
It is also for testing my first commit. 
I will actively contribute.


fassadlr (approved):

fassadlr:
Thanks @lisa3907 !

PR #374, opened 2021-01-14
Expose SubtractFeeFromAmount flag in build-transaction endpoint

zeptin:


src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -417,6 +417,12 @@ public class RecipientModel
         /// </summary>
         public string DestinationScript { get; set; }
 
+        /// <summary>
+        /// An indicator if the transaction fee should be subtracted from the current recipient.
```
zeptin, line 421:
```suggestion
        /// An indicator whether the transaction fee should be subtracted from the current recipient.
```

Restart CI

fassadlr (approved):

PR #404, opened 2021-02-05
Update MaxStandardSigOpsCost and CCTS Max Partials Txs

fassadlr:


src/Stratis.Features.FederatedPeg/TargetChain/CrossChainTransferStore.cs
```diff
@@ -1476,25 +1477,9 @@ private void AddComponentStats(StringBuilder benchLog)
                 benchLog.AppendLine("Failed to retrieve data");
                 this.logger.Error("Exception occurred while getting pending withdrawals: '{0}'.", exception.ToString());
             }
-
-            // Only display this on the mainchain for now as the sidechain has too many SeenInBlock
-            // deposits making the query exectute too long.
-            // TODO: We need to look at including the block height in the dictionary perhaps.
-            if (this.settings.IsMainChain)
```
fassadlr, line 1483: Pretty soon the STRAX side will have too many withdrawals for this code to be appropriate. The "query" is very inefficient.. we don't actually limit the result set.. only after the fact.

We can always get the completed withdrawals via the API controller.

zeptin (approved):

PR #453, opened 2021-03-09
[MulitiSig] Add check to see if one multisig federation member is banning another

fassadlr:
Add this to `api/federationgateway/member/info`


noescape00 (approved):
Looks good

PR #456, opened 2021-03-10
[Cirrus] Create Developer version of CirrusMinerD

fassadlr:
The node needs to start with `-devmode`

Any suggestion on how this can be done better is welcome!

### DIFF
--- a/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
@@ -76,7 +76,7 @@ namespace Stratis.Bitcoin.Features.PoA
 
         private readonly VotingDataEncoder votingDataEncoder;
 
-        private readonly PoAMinerSettings settings;
+        private readonly PoASettings settings;
 
         private readonly IAsyncProvider asyncProvider;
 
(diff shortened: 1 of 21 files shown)


src/Stratis.Bitcoin.Features.PoA/PoAMinerSettings.cs
```diff
@@ -8,6 +8,8 @@ public class PoAMinerSettings
         /// <summary>Allows mining in case node is in IBD and not connected to anyone.</summary>
         public bool BootstrappingMode { get; private set; }
 
+        public bool DevMode { get; private set; }
```
noescape00, line 11: please add comment on why we need dev mode and what it will be used for

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -192,7 +192,7 @@ private async Task CreateBlocksAsync()
                     // The purpose of bootstrap mode is to kickstart the network when the last mined block is very old, which would normally put the node in IBD and inhibit mining.
                     // There is therefore no point keeping this mode enabled once this node has mined successfully.
                     // Additionally, keeping it enabled may result in network splits if this node becomes disconnected from its peers for a prolonged period.
-                    if (this.settings.BootstrappingMode)
+                    if (this.settings.BootstrappingMode && !this.settings.DevMode)
```
noescape00, line 195: How do you bootstrap network when using dev mode?

fassadlr: @StratisIain informed me that idea here is the node should be able to mine and validate smart contracts without it being connected to any other nodes.

src/Stratis.Bitcoin.Features.PoA/PoAMinerSettings.cs
fassadlr: Done.

noescape00 (approved):

PR #497, opened 2021-04-01
[Cirrus] Add ability to have a non-miner join the devmode network

fassadlr:


src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
```diff
@@ -136,7 +136,9 @@ public override Task InitializeAsync()
             if (rebuildFederationHeight)
                 this.reconstructFederationService.Reconstruct();
 
-            this.miner?.InitializeMining();
+            // Initialize mining if the node is started in dev mode and is of role miner.
```
zeptin, line 139: This comment is a bit misleading, it implies that mining should only be started if that is the case

zeptin (approved):

PR #503, opened 2021-04-06
Interflux #0

noescape00:
This can be reviewed on commit by commit basis. 

Mostly refactoring, comments, minor fixes, setup for future adding chains other than ETH for interoperability.


src/Stratis.Features.FederatedPeg/Conversion/ConversionRequest.cs
```diff
@@ -58,6 +67,9 @@ public class ConversionRequest : IBitcoinSerializable
         /// </summary>
         public string DestinationAddress { get { return this.destinationAddress; } set { this.destinationAddress = value; } }
 
+        /// <summary>Chain on which STRAX minting or burning should occur.</summary>
+        public DestinationChain DestinationChain { get { return (DestinationChain)this.destinationChain; } set { this.destinationChain = (int)value; } }
```
zeptin, line 71: If we are introducing this new field we will ideally need a migration mechanism for existing conversion requests stored by the multisig nodes.

noescape00: Given that we now have almost no conversion requests and that we can have multisig nodes upgraded in a few hours I think we can do without it at this stage. 

But I agree, migration mechanism will be needed later on

src/Stratis.Features.FederatedPeg/SourceChain/Deposit.cs
```diff
@@ -12,6 +13,7 @@ public Deposit(uint256 id, DepositRetrievalType retrievalType, Money amount, str
             this.RetrievalType = retrievalType;
             this.Amount = amount;
             this.TargetAddress = targetAddress;
+            //this.TargetChain = targetChain;
```
zeptin, line 16: Is this going to be introduced later?

noescape00: Yes. Adding it in the constructor breaks a lot of tests, I was about to work on it but had to switch to other task, so disabled it in this PR.

src/Stratis.Features.FederatedPeg/TargetChain/MaturedBlocksSyncManager.cs
```diff
@@ -19,6 +19,8 @@ namespace Stratis.Features.FederatedPeg.TargetChain
     /// Handles block syncing between gateways on 2 chains. This node will request
     /// blocks from another chain to look for cross chain deposit transactions.
     /// </summary>
+    /// <remarks>Processes matured block deposits from the cirrus chain and creates instances of <see cref="ConversionRequest"/> which are
+    /// saved to <see cref="IConversionRequestRepository"/>.</remarks>
```
zeptin, line 23: Bear in mind that the creation of `ConversionRequest`s is something that was added onto this existing functionality. Conversion requests are obviously not used for normal cross-chain transfers.

zeptin (approved):

PR #506, opened 2021-04-08
[SC] Contract libraries (proposal)

rowandh:
There's an architectural flaw in the original smart contracts design. Functionality such as hashing, serialization and now ecrecover are exposed via methods on the base `SmartContract` class.

This is not easily extensible as the `SmartContracts` class must be updated every time new functionality is added. Additionally it pollutes the `SmartContracts` class with unnecessary functionality.

This PR adds the beginning of a Smart Contract Library `Stratis.SCL`. The library is whitelisted for use in contracts. To add additional functionality, only the library module needs to be updated.

I've added EcRecover to demonstrate the proposed idea.


rowandh:
@mrtpain would be great to hear your thoughts on this!

rowandh:
If this is accepted then I will update #414 and stratisproject/Stratis.SmartContracts#14 and revert stratisproject/Stratis.SmartContracts#13

YakupIpek:
It is possible to implement these features inside main library without modifying SmartContract class like you do with static classes. Also sometimes modification is unavoidable like we supported uint128/uint256 types and we needed to modify persistent state and other classes. 

In some cases we may need to update both library if we add another library. IMO it would be better to go with single library dependency because i don't see clear benefits.

I am open to hear other ideas .

rowandh:
Thanks @YakupIpek. Definitely some modification is unavoidable. The goal of this is mainly so we can add additional "library-style" functionality more easily in the future. Eg. if we need more support for manipulating byte arrays or converting an enum to a byte 😃 

Are you just suggesting that this functionality be moved to `Stratis.SmartContracts`? The original goal of the `Stratis.SmartContracts` package was just to be a very basic contract interface that rarely needed updating. It contains almost no logic. 

My preference is therefore to separate concerns and create a new namespace for this and add it to the whitelist.

mrtpain:
I think this is a good idea, it'll help keep the SmartContract class clean and let developers choose which added functionality they want to bring into their contracts or keep out entirely.

YakupIpek:
> The goal of this is mainly so we can add additional "library-style" functionality more easily in the future.
How this can make it easier to add new functionality compared to add it to `Stratis.SmartContracts` ? 

I only see separation concerns on this. 

Adding another library bring more work. If i update Stratis.SCL library then i have to update `Stratis.SmartContracts` references again and create pr for it. Already at the moment we have to update Fullnode, SCT tool, Local-Chain project, visual studio template project, sometimes wallet UI..etc and this bring one more project to update process so for this reason i am favor of single library IMHO. We can add another folder for the libraries as namespace in `Stratis.SmartContracts` library.

I would like to know @quantumagi  idea on this.

This is just my suggestion and it is still up to you @rowandh

rowandh:
One of the goals with this is that we _don't_ have to update `Stratis.SmartContracts` every time we want to expose new functionality to a contract.

For example the EcRecover functionality can be exposed through this library, meaning we don't need a `SmartContracts.EcRecover` method.

src/Stratis.FullNode.sln
```diff
@@ -183,7 +183,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stratis.Patricia.Tests", "S
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stratis.Bitcoin.Features.Interop", "Stratis.Bitcoin.Features.Interop\Stratis.Bitcoin.Features.Interop.csproj", "{3DCC6195-1271-4A12-8B94-E821925D98DC}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.External.Masternodes", "Stratis.External.Masternodes\Stratis.External.Masternodes.csproj", "{F04464B5-9D56-4C9A-9778-27B9D314A296}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Stratis.External.Masternodes", "Stratis.External.Masternodes\Stratis.External.Masternodes.csproj", "{F04464B5-9D56-4C9A-9778-27B9D314A296}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.SCL", "Stratis.SCL\Stratis.SCL.csproj", "{F112021F-BE19-4DE2-98F6-827F8518E7CF}"
```
quantumagi, line 188: This reference should be removed as `Stratis.SCL` should only be used by smart contracts.

src/Stratis.SmartContracts.CLR/Stratis.SmartContracts.CLR.csproj
```diff
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Stratis.SCL\Stratis.SCL.csproj" />
```
quantumagi, line 20: This reference should be removed as `Stratis.SCL` should only be used by smart contracts.

quantumagi:
The main benefit of this library is that, if done in a certain way, it could avoid scenarios such as: Package A uses B which is not a package, and `Stratis.SCL` is used by B, then we may have to bump the version of A if `Stratis.SCL` changes. 

The way to accomplish this is to only ever the approproate `Stratis.SCL` library from smart contracts.

src/Stratis.FullNode.sln
rowandh: This isn't a reference, it's just the project in the solution file.

src/Stratis.SmartContracts.CLR/Stratis.SmartContracts.CLR.csproj
rowandh: Unused reference, I have removed it

src/Stratis.SmartContracts.CLR.Validation.Tests/SmartContractDeterminismValidatorTests.cs
```diff
@@ -834,6 +834,44 @@ public Test(ISmartContractState state): base(state)
...
             Assert.NotEmpty(result.Errors);
             Assert.True(result.Errors.All(e => e is ModuleDefinitionValidationResult));
-        }        
+        }
+
+        [Fact]
+        public void SmartContractValidator_Allows_SCL()
+        {
+            var adjustedSource = @"
+using Stratis.SmartContracts;
```
quantumagi, line 843: Can this be loaded directly from the contract file instead?

src/Stratis.SmartContracts.CLR.Validation/FormatPolicy.cs
```diff
@@ -23,7 +23,8 @@ public static class FormatPolicy
             Core,
             typeof(SmartContract).Assembly,
             typeof(Enumerable).Assembly,
-            typeof(IStandardToken).Assembly
+            typeof(IStandardToken).Assembly,
+            typeof(SCL.Base.Operations).Assembly
```
quantumagi, line 27: Should `SCL.Crypto.ECRecover` be added here?

src/Stratis.SmartContracts.CLR/Compilation/ReferencedAssemblyResolver.cs
```diff
@@ -23,7 +23,8 @@ public static class ReferencedAssemblyResolver
                 Core, 
                 typeof(SmartContract).Assembly, 
                 typeof(Enumerable).Assembly,
-                typeof(IStandardToken).Assembly
+                typeof(IStandardToken).Assembly,
+                typeof(SCL.Base.Operations).Assembly
```
quantumagi, line 27: See previous comment.

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/EcRecoverContract.cs
```diff
@@ -0,0 +1,28 @@
...
+            this.State.SetAddress(nameof(this.ThirdPartySigner), value);
+        }
+    }
+
+    public EcRecoverContract(ISmartContractState state, Address thirdPartySigner) : base(state)
+    {
+        this.ThirdPartySigner = thirdPartySigner;
+    }
+
+    public bool CheckThirdPartySignature(byte[] message, byte[] signature)
```
quantumagi, line 23: I am wondering how the signatures used here are being obtained? Since Keccak is being used the normal "signmessage" API would not be the source of these signatures (which uses sha256).

quantumagi (approved):
LGTM. Just a few minor comments.

quantumagi:
image

src/Stratis.SmartContracts.CLR/Stratis.SmartContracts.CLR.csproj
quantumagi: The reference still exits implicitly.

src/Stratis.SmartContracts.CLR.Validation/Stratis.SmartContracts.CLR.Validation.csproj
```diff
@@ -13,4 +13,8 @@
     <PackageReference Include="Stratis.SmartContracts.Standards" Version="2.0.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\Stratis.SCL\Stratis.SCL.csproj" />
```
quantumagi, line 17: Should not have this reference.

quantumagi:
Please ensure that all packages that have been changed or that have project reference to other packages that have updated versions have their versions bumped.

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/EcRecoverContract.cs
rowandh: I copied this verbatim from #414 which was approved. Can remove if you like as it's not essential to this PR

src/Stratis.SmartContracts.CLR.Validation.Tests/SmartContractDeterminismValidatorTests.cs
rowandh: I find it easier to understand the test if you can see the contract at the same time. I know that's a bit inconsistent across the whole solution.

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/EcRecoverContract.cs
quantumagi: @YakupIpek , can you comment on this?

src/Stratis.SmartContracts.CLR.Validation/Stratis.SmartContracts.CLR.Validation.csproj
rowandh: The validator needs to access types from it.

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/EcRecoverContract.cs
rowandh: There's a test where you can see how the signatures are obtained. https://github.com/stratisproject/StratisFullNode/pull/506/files/d4d66b1e2096c7240e9835aa0eabf2df342dbbf4#diff-541900b04541e4da4ce35357818872e3452b844670115c3b54688c7cc6f8d352R76

src/Stratis.SmartContracts.CLR.Validation/Stratis.SmartContracts.CLR.Validation.csproj
quantumagi: Could it access types by name?

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/EcRecoverContract.cs
quantumagi: It shows a programmatic approach. Was instead wondering how this fits into the user experience.

src/Stratis.SmartContracts.CLR.Validation.Tests/SmartContractDeterminismValidatorTests.cs
quantumagi: ok

quantumagi (approved):

rowandh:
To avoid the various versioning issues noted by @quantumagi, I've moved everything into the `Stratis.SmartContracts.CLR` project. This also guarantees that `SCL` will always be loaded in the assembly load context.

I have also moved validation policy definitions out of `Stratis.SmartContracts.CLR.Validation` and into `Stratis.SmartContracts.CLR` which is where they should have been, and bumped both packages a full version as these are significant API changes.

Lastly I will note that this does not solve the 'library upgradeability problem', which is also present with #414. This just makes it possible to add new library functionality without lumping all functionality on the base `SmartContract` class.

quantumagi (changes requested):
This PR shows 58 files changed. That does not seem right.

rowandh:
Retargeted to infracontracts

quantumagi (approved):

PR #507, opened 2021-04-08
[MultiSig] Ensure that input consolidation tx is always built.

fassadlr:
Currently, the multisig is going into a build input consolidation loop. This is the scenario:

1. An withdrawal tx is built that has more than 50 inputs
2. Input consolidation gets triggered
3. Input consolidation "fails" cause of the threshold amount being reached within the first loop already:

`if (oneRound.Sum(x => x.Transaction.Amount) >= amount + FederatedPegSettings.ConsolidationFee)`

4. A consolidation tx is never built due to the above.
5. The method exits and we go back to point 1.
6. The node is now stuck in a infinite loop of trying to build the consolidation tx.

The fix here is that we always need to build the consolidation tx first and THEN check the threshold amount.


noescape00 (approved):
LGTM

quantumagi (approved):
LGTM

zeptin (approved):

PR #511, opened 2021-04-13
Fix errors on MN Launch script.

TjadenFroyda:
Two fixes:
- https://showcase.api.linx.twenty57.net/UnixTime/tounix?date=now changed API to return the unixtimestamp wrapped in quotes. This removes the quotes and gets rid of that error.
- $RemoteTime1 was used before reference; no median was actually taken.


StratisIain:
Nice - Thanks @TjadenFroyda

StratisIain (approved):
Checking UTC Time Difference (unixtime.co.za)
Checking UTC Time Difference (unixtimestamp.com)
SUCCESS: Time difference is 0.229804515838623 seconds

PR #516, opened 2021-04-15
Update Swagger to 6.1.2

rowandh:
Update to latest version of swagger, necessary for future work to add dynamic contract endpoint.

This version removes Json.NET as the default JSON serializer, so I've added it back in and updated the way that string enum serialization is configured.


fassadlr (approved):

rowandh:
The swagger enum serialization here was only used to serialize enums to strings for display in swagger, and not as part of the json API. I believe it's safe to remove because there are no enums in our models.

PR #517, opened 2021-04-15
Fix locks in RocksDB repositories

fassadlr:
The rocks DB implementation of the repos was done by opening and closing the db connection as and when access were required which seems to slow down the sync massively. It is now in line with how the level db repos wortk by opening the connection on start up.


fassadlr:
I have tested a full sync of Strax Mainnet

fassadlr:
I have done:

1. Full syncs on Strax Test / Cirrus Test
2. Staking on Strax Test
3. Mined in Cirrus Test

PR #520, opened 2021-04-15
[MultiSig] Exclude dust from input selection in the MS wallet.

fassadlr:
If we are filtering dust UTXOs from the withdrawal builder we need to also filter them when validating the transaction.

Not doing so fails the logic which checks if earliest unspent UTXO  was used.


quantumagi (approved):
LGTM

For the next PR: IsSpendable is not use in that many places. Should be easy to verify the approach and move the dust check there. That way we don't accumulate more and more dust that will be returned by the index and have to then be "scanned"

PR #526, opened 2021-04-20
[Masternode] Offline signing endpoints

rowandh:
Add two new endpoints 
* Build the join message for signing
* Broadcast the join transaction with a signed message


fassadlr:
Hi Rowan, what was the requirement for this? (Looks good though)

rowandh:
Nothing formal AFAIK, just thought it might be more useful to split the method up this way to enable an offline collateral address to join.

src/Stratis.Bitcoin.Features.PoA/Voting/JoinFederationRequestService.cs
```diff
@@ -92,16 +72,76 @@ public async Task<PubKey> JoinFederationAsync(JoinFederationRequestModel request
                 throw new Exception($"The call to sign the join federation request failed: '{err.Message}'.");
             }
 
+            if (!VerifyCollateralSignature(joinRequest))
```
zeptin, line 75: You check this inside `BroadcastSignedJoinRequestAsync` too - as that method gets called in two places it makes sense to keep the `if` inside it only

src/Stratis.Features.Collateral/CollateralController.cs
```diff
@@ -82,5 +84,97 @@ public async Task<IActionResult> JoinFederationAsync([FromBody] JoinFederationRe
                 return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, e.Message, e.ToString());
             }
         }
+
+        /// <summary>
+        /// Called by a miner wanting to join the federation.
+        /// </summary>
+        /// <param name="request">See <see cref="JoinFederationRequestModel"></see>.</param>
```
zeptin, line 91: Copy pasted?

src/Stratis.Features.Collateral/CollateralController.cs
```diff
@@ -82,5 +84,97 @@ public async Task<IActionResult> JoinFederationAsync([FromBody] JoinFederationRe
...
+                this.logger.Error("Exception occurred: {0}", e.ToString());
+                this.logger.Trace("(-)[ERROR]");
+                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, e.Message, e.ToString());
+            }
+        }
+
+        /// <summary>
+        /// Called by a miner wanting to join the federation.
+        /// </summary>
+        /// <param name="request">See <see cref="JoinFederationRequestModel"></see>.</param>
```
zeptin, line 132: `JoinFederationRequestWithSignature`?

PR #527, opened 2021-04-21
Contract dynamic swagger endpoint

rowandh:
Adds the ability to dynamically generate a Swagger doc for a deployed contract.

Usage:
The contract must first be added to Swagger's definitions by POSTing the a valid contract address to `/swagger/contracts/`. Afterwards, refresh the page to see the contract in the Swagger UI dropdown.

image

POST endpoints - these will create and broadcast a transaction to the network, incurring fees
GET endpoints - these will execute locally and won't incur any fees

Future work:
We could potentially wait for the tx to be mined and return the receipt when making POSTs


fassadlr (approved):
@quantumagi not sure if you wanted to have a look as well?

PR #539, opened 2021-05-03
Add delete chainstate endpoint

rowandh:
Adds the ability to schedule the chain state folders for deletion upon graceful shutdown. This is currently used as a solution to common syncing problems.


fassadlr (approved):
NICE

PR #543, opened 2021-05-05
Rename ChainIndexer GetHeader overloads

rowandh:
The current overloads use different lookup dictionaries. If `GetHeader` is inadvertently called with a ulong it will perform a lookup by hash rather than height due to implicit conversion to a unit256. I made this mistake already.


fassadlr (dismissed):

fassadlr:
@rowandh can you fix the merge conflicts?

PR #559, opened 2021-05-18
[Base] Add dispose last flag

fassadlr:
This is so that we can, for instance, dispose the API last.

The UI client will then be able to call the API right up until the last feature is disposed.


noescape00:
What happens if component is disposed but UI client calls API that uses that component? I think that might lead to crashes in some cases.

fassadlr:
Closing as another solution has been implemented in the UI.

https://github.com/stratisproject/StraxUI/pull/69

PR #560, opened 2021-05-18
[Repos] Move init logic to initialize method

fassadlr:
I still needed to do this for the coindbs. (see #558 )


noescape00 (approved):
LGTM

PR #562, opened 2021-05-20
Add endpoint to get UTXO set for an address

rowandh:
Uses `UtxoIndexer` which is slow in its current implementation because it has to enumerate the whole chain. Improvements could by made by optimizing the indexer.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
@@ -310,7 +311,7 @@ namespace Stratis.Bitcoin.Features.BlockStore.Controllers
             // Get coinview at current height (SLOW)
             var coinView = this.utxoIndexer.GetCoinviewAtHeight(this.chainIndexer.Height);
 
-            var utxos = new List<UtxoModel>();
+            var utxos = new ConcurrentBag<UtxoModel>();
 
             // Get utxos for this address
             try
--- a/src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -289,6 +296,64 @@ public IActionResult GetUtxoSet(int atBlockHeight)
...
+
+                    var utxoModel = new UtxoModel()
+                    {
+                        TxId = utxo.Hash,
+                        Index = utxo.N,
+                        ScriptPubKey = output.ScriptPubKey,
+                        Value = output.Value
+                    };
+
+                    utxos.Add(utxoModel);
```
zeptin, line 339: I think we have to be careful about thread safety here - if the order does indeed not matter then using one of the concurrent collections instead would be preferable

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -289,6 +296,64 @@ public IActionResult GetUtxoSet(int atBlockHeight)
...
+            // Get coinview at current height (SLOW)
+            var coinView = this.utxoIndexer.GetCoinviewAtHeight(this.chainIndexer.Height);
+
+            var utxos = new List<UtxoModel>();
+
+            // Get utxos for this address
+            try
+            {
+                // Fine to do this in parallel because we don't care about the order
+                Parallel.ForEach(coinView.UnspentOutputs, (utxo) =>
```
mrtpain, line 319: Would it be possible/more performance to break this loop when the amount needed for a transaction is found? 

Example: I need uxtos to cover 50 CRS, only return UXTOs needed for up to that amount (maybe plus 1 for gas).

Maybe not an option or worth it, just a thought. Most users of this endpoint will only need _enough_ to cover the transaction not necessary _all_ of them.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
rowandh: Good call, changed to use a concurrent bag.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
rowandh: Definitely. That sounds like a UTXO selection endpoint which isn't the goal with this PR, but could be added after the utxo indexer has been optimized.

zeptin (approved):

PR #573, opened 2021-06-08
[Multsig] Add delete transfer endpoint

fassadlr:
We currently have suspended dynamic conversion transactions sitting on CirrusTest which cannot be reprocessed, so we need the ability to delete them so that the CCTS tip can progress.


quantumagi (approved):
LGTM

PR #584, opened 2021-06-17
Update Dockerfile

fr0stnk:
Changed name of the variable from VERSION to APPVERSION to fix run error (this variable sets system variable and it used by dotnet) and also changed CMD to ENTRYPOINT to give an opportunity to append arguments to docker run


noescape00 (approved):
LGTM

PR #595, opened 2021-06-28
[SC] State.Clear fix

### LINKED ISSUE #594: [SC] State.Clear doesn't work correctly
When `State.Clear` is called it pushes null to the cache. If nothing attempts to read the value before the transaction is executed, the cache value is pushed to storage and everything is fine. However if the value is read before execution is completed, the value in the cache won't have been cleared.

rowandh:
Closes #594.

Of the [currently deployed mainnet contracts](https://github.com/stratisproject/CirrusSmartContracts/tree/master/Mainnet):
* AddressMapper uses clear but doesn't attempt to read the value after clearing and is unaffected
* DaoContract doesn't use clear
* IdentityProvider uses clear but is unaffected
* PrivateYesNoVote doesn't use clear
* StandardToken doesn't use clear

[Testnet](https://github.com/stratisproject/CirrusSmartContracts/tree/master/Testnet):
* Airdrop unaffected
* ICOContract **affected**, currently syncing testnet to check if there will be consensus issues
* StratisSwap unaffected
* TicketBooth unaffected
* WheelGame unaffected

### DIFF
--- a/src/Stratis.SmartContracts.IntegrationTests/SmartContracts/ClearDataContract.cs
+++ b/src/Stratis.SmartContracts.IntegrationTests/SmartContracts/ClearDataContract.cs
@@ -1,0 +1,56 @@
+﻿using Stratis.SmartContracts;
+
+public class ClearDataContract : SmartContract
+{
+    public ClearDataContract(ISmartContractState contractState) : base(contractState)
+    {
+        SetData("Test");
+    }
+
+    public byte[] Data
+    {
+        get
+        {
+            return this.State.GetBytes(nameof(Data));
+        }
+
+        private set
+        {
+            this.State.SetBytes(nameof(Data), value);
+        }
+    }
+
+    public void ClearData()
+    {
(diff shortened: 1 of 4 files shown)


rowandh:
I was able to sync both testnet and mainnet

src/Stratis.SmartContracts.IntegrationTests/SmartContracts/ClearDataContract.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public bool Check()
+    {
+        Assert(Serializer.ToString(this.Data) != "Test", "Value is still test!");
+
+        return this.Data.Length == 0;
+    }
+
+    public void SetData(string data)
```
YakupIpek, line 42: Are these necessary to have ? Already Data property has getter and setters.

YakupIpek (approved):
I tested it on my local chain and it works.

fassadlr (approved):

PR #597, opened 2021-06-28
SmartContract methods for UnityAPI

noescape00:


src/Stratis.Features.Unity3dApi/Controllers/Unity3dController.cs
```diff
@@ -406,5 +441,101 @@ public TipModel GetTip()
...
+            }
+            else
+            {
+                var deserializer = new ApiLogDeserializer(this.primitiveSerializer, this.network, this.stateRoot, this.contractAssemblyCache);
+                List<LogResponse> logResponses = deserializer.MapLogResponses(receipt.Logs);
+
+                receiptResponse = new ReceiptResponse(receipt, logResponses, this.network);
+            }
+            
+            return receiptResponse;
```
fassadlr, line 480: Looks good Ivan, I would just add this controller method and the one below in a try catch, like we do in all the other controller methods. Just so that there is some consistency? WDYT?

Example:

https://github.com/stratisproject/StratisFullNode/blob/master/src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs#L100

fassadlr (approved):
Just add the try catch as in the comment :)

src/Stratis.Features.Unity3dApi/Controllers/Unity3dController.cs
noescape00: There is no need for try catch since if there is an exception it will be seen in node's logs. 
Also we can't use same error handling since controller's methods return explicit types, not json like our other controllers.

PR #601, opened 2021-07-05
[Interflux] Add Ethereum address check to InterFluxOpReturnEncoder

fassadlr:
Further to discussions between me and @zeptin we decided to add back the Eth address validation to the Interflux encoder / deposit extraction.


src/Stratis.Bitcoin.Features.Wallet/InterFluxOpReturnEncoder.cs
```diff
@@ -41,4 +62,16 @@ public static bool TryDecode(byte[] opReturnData, out int destinationChain, out
             return TryDecode(stringData, out destinationChain, out address);
         }
     }
+
+    /// <summary>Chains supported by InterFlux integration.</summary>
+    public enum DestinationChain
+    {
+        STRAX = 0, // Stratis
+        ETH, // Ethereum
```
zeptin, line 70: It may be wise to explicitly set all the rest, so that no one is tempted to insert new chains in the middle of the enum in future

zeptin (approved):
Looks good to me

noescape00 (approved):
LGTM

PR #608, opened 2021-07-08
[InterFlux] Only allow ETH Interflux conversions

fassadlr:
Currently, only ETH conversion transactions are fully tested by the multisig, so for the time being we should only allow these interflux conversions.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
@@ -442,13 +442,14 @@ namespace Stratis.Bitcoin.Features.Wallet.Controllers
         [ProducesResponseType((int)HttpStatusCode.OK)]
         [ProducesResponseType((int)HttpStatusCode.BadRequest)]
         [ProducesResponseType((int)HttpStatusCode.InternalServerError)]
-        public async Task<IActionResult> BuildInterFluxTransaction([FromBody] BuildInterFluxTransactionRequest request,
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -442,13 +442,14 @@ public IActionResult GetHistory([FromQuery] WalletHistoryRequest request)
         [ProducesResponseType((int)HttpStatusCode.OK)]
         [ProducesResponseType((int)HttpStatusCode.BadRequest)]
         [ProducesResponseType((int)HttpStatusCode.InternalServerError)]
-        public async Task<IActionResult> BuildInterFluxTransaction([FromBody] BuildInterFluxTransactionRequest request,
-            CancellationToken cancellationToken = default(CancellationToken))
+        public async Task<IActionResult> BuildInterFluxTransaction([FromBody] BuildInterFluxTransactionRequest request)
         {
+            if (request.DestinationChain != (int)DestinationChain.ETH)
+                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, "Invalid destination chain", "Only InterFlux transactions to the Ethereum chain is currently supported.");
```
zeptin, line 448:
```suggestion
                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, "Invalid destination chain", "Only InterFlux transactions to the Ethereum chain are currently supported.");
```

zeptin (approved):

PR #616, opened 2021-07-09
[InterFlux] Remove confirmation loop for conversion transaction submission

fassadlr:


src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -425,8 +425,11 @@ private async Task ProcessConversionRequestsAsync()
 
                             request.RequestEthTransactionHash = identifiers.TransactionHash;
 
-                            if (!await WaitForTransactionToBeConfirmedAsync(identifiers, request.DestinationChain).ConfigureAwait(false))
-                                return;
+                            (BigInteger confirmationCount, string blockHash) = await this.ethClientProvider.GetClientForChain(request.DestinationChain).GetConfirmationsAsync(identifiers.TransactionHash).ConfigureAwait(false);
+                            this.logger.LogInformation($"Originator confirming transaction id '{identifiers.TransactionHash}' '({identifiers.TransactionId})' before broadcasting; confirmations: {confirmationCount}; Block Hash {blockHash}.");
+
+                            if (confirmationCount < this.SubmissionConfirmationThreshold)
+                                break;
```
zeptin, line 432: Hmm, is it safe to break without changing the state? Given that the transaction has been submitted even if it is insufficiently confirmed

zeptin (approved):

PR #625, opened 2021-07-15
[InterFlux] Move burn request processing to deposit extractor

fassadlr:


src/Stratis.Features.FederatedPeg/SourceChain/DepositExtractor.cs
```diff
@@ -73,6 +76,8 @@ public IReadOnlyList<IDeposit> ExtractDepositsFromBlock(Block block, int blockHe
             else
                 deposits = new List<IDeposit>();
 
+            ProcessInterFluxBurnRequests(deposits, blockHeight);
```
zeptin, line 79: Hmm, presumably this should only run on the Cirrus side

zeptin (approved):

Issue #661, opened 2021-08-11
Local Call Log Deserialization


mrtpain:
As an API user, I would like to have Local Call logs deserialized before the request is returned from `SmartContracts/local-call`. Currently the `internalTransfers` do not output the address values and log `topics` and `data` are not returning deserialized values as they do at the `SmartContracts/receipt` endpoint. 

Using master branch 60cff14fbcd050e02bad0c4e5dbbefb26ba41018

**Current Response**
```
{
    "internalTransfers": [
        {
            "from": {
                "size": 20
            },
            "to": {
                "size": 20
            },
            "value": 0
        }
    ],
    "gasConsumed": {
        "value": 33810
    },
    "revert": false,
    "errorMessage": null,
    "return": null,
    "logs": [
        {
            "address": {
                "size": 20
            },
            "topics": [
                "VHJhbnNmZXJMb2c=",
                "FNys6Fs53wRCRQyDlaFu5ODb+I8=",
                "ic5TcYHmoKaTwsCEUOOsw9kqxfg="
            ],
            "data": "+EuUFNys6Fs53wRCRQyDlaFu5ODb+I+Uic5TcYHmoKaTwsCEUOOsw9kqxfigECcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
        },
        {
            "address": {
                "size": 20
            },
            "topics": [
                "U3RhcnRNaW5pbmdMb2c=",
                "FNys6Fs53wRCRQyDlaFu5ODb+I8="
            ],
            "data": "+HiUFNys6Fs53wRCRQyDlaFu5ODb+I+gECcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgUOKjcuEVjQUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgUOJU5qwteAMAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
        }
    ]
}
```

**Expected Response**
```
{
    "internalTransfers": [
        {
            "from": "PAVV2c9Muk9Eu4wi8Fqdmm55ffzhAFPffV",
            "to": "PM9pu2ppUaRU1WMsUnq5en1b4ruwLqqEpX",
            "value": 0
        }
    ],
    "gasConsumed": {
        "value": 33810
    },
    "revert": false,
    "errorMessage": null,
    "return": null,
    "logs": [
        {
[...]

PR #662, opened 2021-08-12
[SC] Local call log serialization

rowandh:
See #661.

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts.Tests/ApiLogDeserializerTests.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts.Tests/ApiLogDeserializerTests.cs
@@ -11,6 +11,7 @@ using Stratis.SmartContracts.CLR.Loader;
 using Stratis.SmartContracts.CLR.Serialization;
 using Stratis.SmartContracts.Core.Receipts;
 using Stratis.SmartContracts.Core.State;
+using Stratis.SmartContracts.Core.State.AccountAbstractionLayer;
 using Stratis.SmartContracts.Networks;
 using Xunit;
 
@@ -112,6 +113,36 @@ namespace Stratis.Bitcoin.Features.SmartContracts.Tests
             // Verify that we got the code for both log assemblies.
             stateRoot.Verify(s => s.GetCodeHash(logs[0].Address), Times.Once);
             stateRoot.Verify(s => s.GetCodeHash(logs[1].Address), Times.Once);
+        }
+
+        [Fact]
+        public void MapTransferInfo_Success()
+        {
+            var network = new SmartContractsRegTest();
+            var primitiveSerializer = new ContractPrimitiveSerializer(network);
+
+            var stateRoot = new Mock<IStateRepositoryRoot>();
+            stateRoot.Setup(r => r.GetCodeHash(It.IsAny<uint160>())).Returns(uint256.Zero.ToBytes());
+
+            var assemblyCache = new Mock<IContractAssemblyCache>();
+            var contractAssembly = new Mock<IContractAssembly>();
+
+            var serializer = new ApiLogDeserializer(primitiveSerializer, network, stateRoot.Object, assemblyCache.Object);
+
+            var transferInfos = new List<TransferInfo>
+            {
+                new TransferInfo(uint160.Zero, uint160.One, 12345),
+                new TransferInfo(uint160.One, uint160.Zero, 12345),
+            };
+
+            List<SmartContracts.Models.TransferResponse> result = serializer.MapTransferInfo(transferInfos.ToArray());
+
+            for (var i = 0; i < transferInfos.Count; i++)
+            {
+                Assert.Equal(transferInfos[i].From.ToBase58Address(network), result[i].From);
+                Assert.Equal(transferInfos[i].To.ToBase58Address(network), result[i].To);
+                Assert.Equal(transferInfos[i].Value, result[i].Value);
+            }
         }
     }
 }
--- a/src/Stratis.Bitcoin.Features.SmartContracts.Tests/ApiLogDeserializerTests.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts.Tests/ApiLogDeserializerTests.cs
@@ -4,6 +4,7 @@ using System.Reflection;
 using System.Text;
 using Moq;
 using NBitcoin;
+using Stratis.Bitcoin.Features.SmartContracts.Models;
 using Stratis.SmartContracts;
 using Stratis.SmartContracts.CLR;
 using Stratis.SmartContracts.CLR.Caching;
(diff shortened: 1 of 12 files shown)


rowandh:
@mrtpain would appreciate your input into whether this resolves #661!

fassadlr:
@rowandh it looks good. Just need @mrtpain to comment 👍 

Also, I see the windows build was cancelled?

rowandh:
Strange, looks like the build died. Will kick it off again.

fassadlr (approved):
Merge if CI passes

rowandh:
@mrtpain the return value should be fixed now, could you please check once more and confirm?

mrtpain:
@rowandh 🥳 thanks all is good now 👍

mrtpain:
Actually, seems Addresses are returning `{}` from the `return` property.

rowandh:
@mrtpain should be fixed now

mrtpain:
Thanks @rowandh all good now. 

Somewhat related: recent versions of FN output the `event` in logs. This is helpful to have so clients don't need to deserialize topics, but I wonder, what would happen if I output a contract log that had an `Event` property already? 

```
"log": {
    "event": "CollectMiningRewardsLog",
    "miner": "PPQdeXdjWDBzVLUjgWwi4mFP4Y1mhuNcRu",
    "amount": "2541009128850074"
}
```

Currently we are deserializing the first topic to get our event names. We can remove a package dependency if we can rely on the `log.event` property but wondering about the potential conflict outlined above.

rowandh:
> Thanks @rowandh all good now.
> 
> Somewhat related: recent versions of FN output the `event` in logs. This is helpful to have so clients don't need to deserialize topics, but I wonder, what would happen if I output a contract log that had an `Event` property already?
> 
> ```
> "log": {
>     "event": "CollectMiningRewardsLog",
>     "miner": "PPQdeXdjWDBzVLUjgWwi4mFP4Y1mhuNcRu",
>     "amount": "2541009128850074"
> }
> ```
> 
> Currently we are deserializing the first topic to get our event names. We can remove a package dependency if we can rely on the `log.event` property but wondering about the potential conflict outlined above.

Good point, a log property named `Event` would replace the event name. We should be nesting the actual log data, something like this:
```
"log": {
    "event": "CollectMiningRewardsLog",
    "data": {
        "miner": "PPQdeXdjWDBzVLUjgWwi4mFP4Y1mhuNcRu",
        "amount": "2541009128850074"
   }
}
```

mrtpain:
On another note, @rowandh, transactions with internal CRS transfers I believe either aren't implemented fully (intentionally) or may have a bug.

We are able to successfully broadcast transactions that return errors during a local call. These errors are pointing to a 0 CRS amount in Message.Value. 

https://github.com/Opdex/opdex-v1-core/blob/main/src/Contracts/Routers/OpdexRouter.cs#L228

That line is throwing during local calls but the same payload sent for an actual transaction is successful. (Note that is a public method but in our scenario of calling `Router.AddLiquidity` it is being called from within other methods).

Along with the error that we receive in the local call response, we can also see the list of internal transfers showing a 0 amount.

drmathias:
@rowandh We're getting a 500 response making a local call to [this method](https://github.com/Opdex/opdex-v1-core/blob/b7775f9fdb00766006bf9f84c057d7d84df112ef/src/Contracts/Deployer/OpdexMarketDeployer.cs#L58-L100), when running the full node built from this branch. The request body is below.

```json
{
  "walletName": "{{wallet_name}}",
  "accountName": "{{wallet_account}}",
  "contractAddress": "{{market_deployer}}",
  "methodName": "CreateStandardMarket",
  "amount": "0",
  "feeAmount": ".001",
  "password": "{{wallet_password}}",
  "gasPrice": 100,
  "gasLimit": 250000,
  "sender": "{{wallet_address}}",
  "parameters": [
    "9#{{wallet_address}}", // market owner
    "5#3", // fee
    "1#True", // authPoolCreators
    "1#True", // authProviders
    "1#False", // authTraders
    "1#True" // enableMarketFee
  ]
}
```

Had a go at debugging locally and it seems to be failing to map the log responses.

YakupIpek:
> @rowandh We're getting a 500 response making a local call to [this method](https://github.com/Opdex/opdex-v1-core/blob/b7775f9fdb00766006bf9f84c057d7d84df112ef/src/Contracts/Deployer/OpdexMarketDeployer.cs#L58-L100), when running the full node built from this branch. The request body is below.
> 
> ```json
> {
>   "walletName": "{{wallet_name}}",
>   "accountName": "{{wallet_account}}",
>   "contractAddress": "{{market_deployer}}",
>   "methodName": "CreateStandardMarket",
>   "amount": "0",
>   "feeAmount": ".001",
>   "password": "{{wallet_password}}",
>   "gasPrice": 100,
>   "gasLimit": 250000,
>   "sender": "{{wallet_address}}",
>   "parameters": [
>     "9#{{wallet_address}}", // market owner
>     "5#3", // fee
>     "1#True", // authPoolCreators
>     "1#True", // authProviders
>     "1#False", // authTraders
>     "1#True" // enableMarketFee
>   ]
> }
> ```
> 
> Had a go at debugging locally and it seems to be failing to map the log responses.

Is this on testnet, mainnet or in localchain(devmode) ? Can you be sure that wallet_address is belong to one of your wallet addresses ?

mrtpain:
> > @rowandh We're getting a 500 response making a local call to [this method](https://github.com/Opdex/opdex-v1-core/blob/b7775f9fdb00766006bf9f84c057d7d84df112ef/src/Contracts/Deployer/OpdexMarketDeployer.cs#L58-L100), when running the full node built from this branch. The request body is below.
> > ```json
> > {
> >   "walletName": "{{wallet_name}}",
> >   "accountName": "{{wallet_account}}",
> >   "contractAddress": "{{market_deployer}}",
> >   "methodName": "CreateStandardMarket",
> >   "amount": "0",
> >   "feeAmount": ".001",
> >   "password": "{{wallet_password}}",
> >   "gasPrice": 100,
> >   "gasLimit": 250000,
> >   "sender": "{{wallet_address}}",
> >   "parameters": [
> >     "9#{{wallet_address}}", // market owner
> >     "5#3", // fee
> >     "1#True", // authPoolCreators
> >     "1#True", // authProviders
> >     "1#False", // authTraders
> >     "1#True" // enableMarketFee
> >   ]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Had a go at debugging locally and it seems to be failing to map the log responses.
> 
> Is this on testnet, mainnet or in localchain(devmode) ? Can you be sure that wallet_address is belong to one of your wallet addresses ?

This is our local network (devnet) running from this branch. Our payload is all correct.

We can switch the endooint from `smartcontracts/local-call` to `smartcontractwallet/call` and it works fine. Also we can change just the parameters and method to `CreateStakingMarket` method with a local call and it works as expected. 

So far it seems that it's a problem with something happening in this method specifically that is uncaught as the FN returns 500 error with no body, I do not think the error is getting caught but throwing all the way out. 

Also local calls that involve CRS transfers and use of `Message.Value` in contract always have it as 0 and the local call tx returns failures due to it.

rowandh:
Will merge these changes in to release/1.0.9.6 and address the new issues with separate PRs.

rowandh:
f9b3e80 resolves zero amount local calls (https://github.com/stratisproject/StratisFullNode/pull/662#issuecomment-902379004)
ae7099c resolves the issue deserializing addresses (https://github.com/stratisproject/StratisFullNode/pull/662#issuecomment-907840219)
f7f278a resolves nesting event data (https://github.com/stratisproject/StratisFullNode/pull/662#issuecomment-899937736)

fassadlr:
@rowandh I believe this fix was merged to release/1.1.1.0... can you confirm? If so it was also then merge to 1.2 already.

PR #666, opened 2021-08-16
Unity3dAPI: added receipt search

noescape00:
Pretty straight forward, it just calls receipt search from SC controller

### DIFF
--- a/src/Stratis.Features.Unity3dApi/Controllers/Unity3dController.cs
+++ b/src/Stratis.Features.Unity3dApi/Controllers/Unity3dController.cs
@@ -17,7 +17,6 @@ using Stratis.Bitcoin.Features.Consensus;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
 using Stratis.Bitcoin.Features.SmartContracts;
 using Stratis.Bitcoin.Features.SmartContracts.Models;
-using Stratis.Bitcoin.Features.SmartContracts.ReflectionExecutor.Controllers;
 using Stratis.Bitcoin.Features.SmartContracts.Wallet;
 using Stratis.Bitcoin.Features.Wallet.Controllers;
 using Stratis.Bitcoin.Features.Wallet.Models;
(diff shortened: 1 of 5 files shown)


src/Stratis.Features.Unity3dApi/Controllers/Unity3dController.cs
```diff
@@ -511,6 +516,32 @@ public LocalExecutionResult LocalCallSmartContractTransaction([FromBody] LocalCa
...
+        /// <param name="topics">The topics to search. All specified topics must be present.</param>
+        /// <param name="fromBlock">The block number from which to start searching.</param>
+        /// <param name="toBlock">The block number where searching finishes.</param>
+        /// 
+        /// <returns>A list of receipts for transactions relating to a specific smart contract and a specific event in that smart contract.</returns>
+        [Route("api/[controller]/receipt-search")]
+        [HttpGet]
+        public async Task<List<ReceiptResponse>> ReceiptSearchAPI([FromQuery] string contractAddress, [FromQuery] string eventName, [FromQuery] List<string> topics = null, [FromQuery] int fromBlock = 0, [FromQuery] int? toBlock = null)
+        {
+            List<ReceiptResponse> result = this.smartContractsController.ReceiptSearch(contractAddress, eventName, topics, fromBlock, toBlock);
```
fassadlr, line 540: Hi Ivan,

It seems that the logic in `SmartContractsController.ReceiptSearch` is now shared between 2 controllers. There is already a `smartContractTransactionService` in this controller, so move the logic in `SmartContractsController.ReceiptSearch` there and then just call `smartContractTransactionService.ReceipSearch` from here (also just update `SmartContractsController` to do the same.

😎

noescape00: `It seems that the logic in SmartContractsController.ReceiptSearch is now shared between 2 controllers.`
Logic was just in SC controller and unity controller called it (same was done with other methods as well)

But anyway, changes applied

fassadlr (approved):
I know its a bit more code bit this looks a lot cleaner now ;) I was worried also that some person randomly moves or even remove the sc controller and then your code breaks.

noescape00:
Closing this one and moving it to https://github.com/stratisproject/StratisFullNode/pull/712

PR #744, opened 2021-10-20
[SC] Fix local call message value zero

rowandh:
See https://github.com/stratisproject/StratisFullNode/pull/662#issuecomment-902379004


src/Stratis.SmartContracts.IntegrationTests/ContractBehaviourTests.cs
```diff
@@ -230,6 +230,72 @@ public void Local_Call_Should_Produce_Logs_And_Transfers()
...
+            Assert.Equal(0UL, this.node1.GetContractBalance(recipientResponse.NewContractAddress));
+
+            ReceiptResponse receipt2 = this.node1.GetReceipt(recipientResponse.TransactionId.ToString());
+            Assert.True(receipt2.Success);
+
+            uint256 currentHash = this.node1.GetLastBlock().GetHash();
+
+            NBitcoin.Block lastBlock = this.node1.GetLastBlock();
+
+            // Create a log in a local call
```
zeptin, line 275:
```suggestion
            // Create a log in a local call.
```
Boot CI

zeptin (approved):

fassadlr:
@rowandh I believe this fix was merged to release/1.1.1.0... can you confirm? If so it was also then merge to 1.2 already.

rowandh:
@fassadlr looks like the fix is in but not the tests, so we should still merge this.

PR #760, opened 2021-11-08
[RPC] Added two RPC methods for listing transactions

fenix2222:
Added implementation for 

listtransactions
listsinceblock

RPC methods

### DIFF
--- a/src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
+++ b/src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
@@ -23,7 +23,7 @@ namespace Stratis.Bitcoin.Features.RPC.Models
             this.Amount = new Money((long)(amount * Money.COIN));
             this.Confirmations = (int)tx["confirmations"];
             this.TxId = (string)tx["txid"];
-            this.Index = (int)tx["blockindex"];
+            this.BlockIndex = (int)tx["blockindex"];
 
             if (tx["generated"] != null)
             {
@@ -42,10 +42,10 @@ namespace Stratis.Bitcoin.Features.RPC.Models
         public string TxId { get; set; }
 
         /// <summary>
-        /// The index of the output in the transaction.
+        /// The index of the transaction in the block that includes it.
         /// </summary>
         [JsonProperty(PropertyName = "blockindex")]
-        public int Index { get; set; }
+        public int BlockIndex { get; set; }
 
         /// <summary>
         /// Only present if transaction only input is a coinbase one.
--- a/src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
+++ b/src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
@@ -1,0 +1,79 @@
+﻿using NBitcoin;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Stratis.Bitcoin.Features.RPC.Models
+{
+    public class TransactionInfoModel
+    {
+        public TransactionInfoModel()
+        {
+
+        }
+
+        public TransactionInfoModel(JObject tx)
+        {
+            if (tx["blockheight"] != null)
+            {
+                this.BlockHeight = (int)tx["blockheight"];
+            }
+
+            this.BlockHash = (string)tx["blockhash"];
+            decimal amount = (decimal)tx["amount"];
+            this.Amount = new Money((long)(amount * Money.COIN));
+            this.Confirmations = (int)tx["confirmations"];
+            this.TxId = (string)tx["txid"];
+            this.Index = (int)tx["blockindex"];
+
+            if (tx["generated"] != null)
+            {
+                this.Generated = (bool)tx["generated"];
+            }
+            else
+            {
+                // Default to True for earlier versions, i.e. if not present
+                this.Generated = false;
+            }
+        }
+
+        /// <summary>
+        /// The transaction id.</summary>
+        [JsonProperty(Order = 1, PropertyName = "txid")]
+        public string TxId { get; set; }
+
+        /// <summary>
+        /// The index of the output in the transaction.
+        /// </summary>
+        [JsonProperty(PropertyName = "blockindex")]
+        public int Index { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


src/Stratis.Bitcoin.IntegrationTests/RPC/RpcBitcoinImmutableTests.cs
```diff
@@ -78,6 +79,22 @@ public async Task GetTxOutAsyncWithValidTxThenReturnsCorrectUnspentTxAsync()
             Assert.Equal(coin.Address.ToString(), resultTxOut.scriptPubKey.addresses[0]);
         }
 
+        [Fact]
+        public async Task GetTxOutAsyncWithValidTxThenReturnsCorrectTxsAsync()
```
zeptin, line 83: The naming of the tests is a bit misleading

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -790,6 +791,75 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
             return unspentCoins.ToArray();
         }
 
+        [ActionName("listtransactions")]
+        [ActionDescription("Returns an array of transactions belonging to this wallet.")]
+        public TransactionInfoModel[] ListTransactions(string label = "*", int count = 10, int skip = 0)
```
zeptin, line 796: Should we have the `include_watchonly` parameter as well? As we do make a distinction between transactions in 'normal' accounts vs special accounts. And it makes sense to make the parameter list match as closely as possible in general.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -790,6 +791,75 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+
+        [ActionName("listsinceblock")]
+        [ActionDescription("Returns an array of transactions belonging to this wallet since specified block.")]
+        public TransactionsSinceBlockModel ListSinceBlock(string blockhash = "", int targetConfirmations = 1)
+        {
+            WalletAccountReference accountReference = this.GetWalletAccountReference();
+            var transactions = this.walletManager.GetAllTransactionsInWallet(accountReference.WalletName);
+
+            int? targetBlockHeight = null;
+            if (string.IsNullOrEmpty(blockhash))
```
zeptin, line 828: Missing the `!` I think

src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+
+        /// <summary>
+        /// The transaction id.</summary>
+        [JsonProperty(Order = 1, PropertyName = "txid")]
+        public string TxId { get; set; }
+
+        /// <summary>
+        /// The index of the output in the transaction.
```
zeptin, line 45: Correction: `The index of the transaction in the block that includes it.`

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1090,6 +1090,20 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactionsInWallet(stri
             return res;
         }
 
+        public IEnumerable<TransactionData> GetAllTransactionsInWallet(string walletName)
```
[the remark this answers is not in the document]
zeptin: That then allows the caller to be more explicit about what they want, rather than relying on the default behaviour of `GetAllTransactions` returning normal account transactions only. We generally would not want the cold staking account transactions to be returned in the wallet RPCs but the watch only account transactions often should be. Cold staking transactions require special care due to their spendability being more complex to determine, so it is safer not to return them.

src/Stratis.Bitcoin.Features.RPC/Models/TransactionInfoModel.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "blockindex")]
+        public int Index { get; set; }
```
zeptin, line 48: Based on what this field is intended to represent, the property should be named `BlockIndex` as well

src/Stratis.Bitcoin.Features.RPC/Models/TransactionsSinceBlockModel.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        /// <summary>
+        /// All transactions.</summary>
+        [JsonProperty(Order = 1, PropertyName = "transactions")]
+        public TransactionInfoModel[] Transactions { get; set; }
+
+        /// <summary>
+        /// The hash of the block since which we got transactions.</summary>
+        [JsonProperty(PropertyName = "lastblock")]
+        public string Lastblock { get; set; }
```
zeptin, line 15: `LastBlock`

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -790,6 +791,75 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+                    .Take(count)
+                    .Select(tx => new TransactionInfoModel
+                    {
+                        BlockHash = tx.BlockHash.ToString(),
+                        Amount = tx.Amount,
+                        TxId = tx.Id.ToString(),
+                        BlockHeight = tx.BlockHeight ?? 0,
+                        Generated = tx.IsCoinBase ?? false,
+                        Confirmations = tx.BlockHeight.HasValue ? (this.walletManager.LastBlockHeight() + 1) - tx.BlockHeight.Value : 0,
+                        Index = tx.Index
```
zeptin, line 813: As mentioned, this is actually supposed to be the position of the transaction within the block, not the index of the UTXO being tracked by the wallet. I think `TransactionData` does include a `BlockIndex` property that can be used?

fenix2222:
Closing this one and moving it to https://github.com/stratisproject/StratisFullNode/pull/764 as it requires 1.0.9.8

PR #784, opened 2021-11-26
[Proposal] Show CS address in getrawtransaction output

zeptin:
The rationale for this is in the code comments; essentially this is to improve the user experience for block explorer users trying to check their cold staking balances or staking activity.

After picking a cold staking transaction ID at random, the modified `getrawtransaction` output looks like this:

```
    {
      "value": 780.5398,
      "n": 1,
      "scriptPubKey": {
        "asm": "OP_DUP OP_HASH160 OP_ROT OP_IF OP_CHECKCOLDSTAKEVERIFY 98f869eb17b7d9ed5529bb8d139d960624ab7c39 OP_ELSE d8e6de0c675ad75f430ccbd08f3a8b0e519dc6ec OP_ENDIF OP_EQUALVERIFY OP_CHECKSIG",
        "hex": "76a97b63b91498f869eb17b7d9ed5529bb8d139d960624ab7c396714d8e6de0c675ad75f430ccbd08f3a8b0e519dc6ec6888ac",
        "reqSigs": 1,
        "type": "coldstaking",
        "addresses": [
          "Xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
        ]
      }
    }
```

### DIFF
--- a/src/Stratis.Bitcoin/Controllers/Models/TransactionModel.cs
+++ b/src/Stratis.Bitcoin/Controllers/Models/TransactionModel.cs
@@ -281,6 +281,8 @@ namespace Stratis.Bitcoin.Controllers.Models
         /// <param name="network">The network where the transaction was conducted.</param>
         public ScriptPubKey(NBitcoin.Script script, Network network) : base(script)
         {
+            this.Type = this.GetScriptType(script.FindTemplate(network));
+
             // To avoid modifying the very low-level GetDestination logic, check for cold staking scripts first.
             // The decision to show the cold pubkey's address in the 'addresses' list is based on the following:
             // 1. It seems more intuitive from a user's perspective that their balance will appear against this address.
@@ -299,7 +301,6 @@ namespace Stratis.Bitcoin.Controllers.Models
             }
 
             var destinations = new List<TxDestination> { script.GetDestination(network) };
-            this.Type = this.GetScriptType(script.FindTemplate(network));
 
             if (destinations[0] == null)
             {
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Controllers/Models/TransactionModel.cs
```diff
@@ -281,9 +281,27 @@ public ScriptPubKey()
         /// <param name="network">The network where the transaction was conducted.</param>
         public ScriptPubKey(NBitcoin.Script script, Network network) : base(script)
         {
-            var destinations = new List<TxDestination> { script.GetDestination(network) };
             this.Type = this.GetScriptType(script.FindTemplate(network));
 
+            // To avoid modifying the very low-level GetDestination logic, check for cold staking scripts first.
```
zeptin, line 286:
```suggestion
            // To avoid modifying the very low-level GetDestination logic, check for a cold staking script first.
```

Boot CI

quantumagi (approved):
LGTM

zeptin:
> LGTM. My only concern is that we should probably not be showing the pubkeys that make up the cold staking script to the user as normal addresses at all - to prevent them from being used for spending funds towards. This represents a move away from that potential goal.

In the interests of making as few changes at a time as possible, perhaps this should go in as-is, but we could indeed change the version byte for cold staking addresses in future. This would require changes to the cold staking controller and possibly the wallet UI too. There may be other places in the codebase that would need to be made aware of a new address type. 

Either that or make the display of the addresses purely cosmetic output with no recognition of the modified addresses in the codebase. This could have its own drawbacks though, as the API/RPC would be outputting 'addresses' that it can't then give any further information about.

PR #814, opened 2021-12-08
Serializer deployments

rowandh:
Adds the updated serializer and an activation height on the network.


fassadlr:
Looks good @rowandh, just set the mainnet height to somewhere later today and we are good 👍

fassadlr (approved):

PR #866, opened 2022-01-28
[InterFlux] Add resetscanheight endpoint

zeptin:
Allows the interop poller to be reset to a specified height for a given chain


fassadlr (approved):
Tight!

PR #867, opened 2022-01-28
Disable price tracking by default

zeptin:
The ExternalApi price tracking polling loops were enabled by default in StraxD; this disables them by default. They can be re-enabled via command line switches for the multisig masternodes as needed.


fassadlr (approved):

fassadlr (changes requested):
Please point to 1.2

PR #868, opened 2022-01-30
NotFound result if key is not found

YakupIpek:


zeptin (approved):
@fassadlr Is it ok for this to go into 1.2.0.0?

mrtpain:
cc @drmathias when this is included in the official 1.2.0.0 package, we can adjust for this change

PR #913, opened 2022-02-28
[InterFlux] Distribute transfer fee to multisig

fassadlr:


src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -430,37 +434,31 @@ private async Task PollCirrusForBurnsAsync(BigInteger blockHeight)
...
-                                    conversionFeeOutput = txOut;
-                                }
-                            }
+                            // Check if the fee and transfer amount is valid.
+                            byte[] valueBytes = src20burn.Value.ToByteArray();
+                            byte[] transferRequestAmountBytes = new byte[32];
+                            Array.Copy(valueBytes, 0, transferRequestAmountBytes, 0, valueBytes.Length);
 
-                            if (conversionFeeOutput == null)
+                            var transferRequestAmount = new BigInteger(transferRequestAmountBytes);
```
fassadlr, line 442: Kevin please review the amounts carefully to ensure that my conversions to the various types are correct.

zeptin, line 442: Isn't `src20burn.Value` already a `BigInteger`? This byte manipulation is needed for creating a `uint256` from a `BigInteger`'s byte array representation, as they zero pad differently

src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -474,8 +472,8 @@ private async Task PollCirrusForBurnsAsync(BigInteger blockHeight)
                                         RequestType = ConversionRequestType.Burn,
                                         Processed = false,
                                         RequestStatus = ConversionRequestStatus.Unprocessed,
-                                        Amount = new uint256(paddedArray),
-                                        BlockHeight = (int)blockHeight,
+                                        Amount = new uint256(transferRequestAmount.ToByteArray()),
```
zeptin, line 475: As mentioned before, this is not valid. You have to prepend a sufficient number of zeros for the uint256 constructor to accept the byte array. BigInteger basically only emits sufficient bytes to represent the internal integer, they aren't aligned to any particular size like uint256 is.

Because BigInteger is somewhat smarter about instantiation, you don't have to do the padding when converting from a uint256 to a BigInteger

src/Stratis.Features.FederatedPeg/TargetChain/MaturedBlocksSyncManager.cs
```diff
@@ -115,6 +128,21 @@ public async Task StartAsync()
             repeatEvery: TimeSpan.FromSeconds(RefreshDelaySeconds));
         }
 
+        /// <inheritdoc />
+        public void AddInterOpFeeDeposit(IDeposit deposit)
+        {
+            lock (this.lockObject)
+            {
+                if (this.interOpFeeDeposits.Select(d => d.Id).Contains(deposit.Id))
```
zeptin, line 136: You can probably use `.Any()` here?

zeptin (approved):

PR #917, opened 2022-03-01
[InterFlux] Move supported contracts controller to Stratis.Bitcoin

fassadlr:


zeptin (approved):

src/Stratis.Interop.Contracts/SupportedContractAddresses.cs
```diff
@@ -0,0 +1,78 @@
...
+        public static List<SupportedContractAddress> ForNetwork(NetworkType networkType)
+        {
+            return contractAddresses.GetValueOrDefault(networkType);
+        }
+    }
+
+    public enum SupportedNativeChain
+    {
+        Ethereum
+    }
```
mrtpain, line 62: Can the JSON Serialization be changed for this so it returns out the string value of the name rather than the enum Id? 

Ideally resulting in `"nativeNetwork": "Ethereum"` rather than `"nativeNetwork": 0`.

Otherwise this LGTM, thanks!

PR #935, opened 2022-03-14
[InterFlux] Allow polling for transfers on Cirrus to also catch up faster when behind

fassadlr:
Currently if a MS node stops for eg. 15 mins and restarts, the Cirrus transfer polling loop will still sync up only one block for each run of the polling loop. This is inconsistent with the behavior when polling for applicable blocks on the ETH chain

### DIFF
--- a/src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
+++ b/src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
@@ -51,7 +51,7 @@ namespace Stratis.Bitcoin.Features.Interop
         /// If the last polled block for a given destination chain is more than this amount of blocks from the chain's tip,
         /// fast sync it up before the async loop poller takes over.
         /// </summary>
-        private const int DestinationChainSyncToBuffer = 25;
+        private const int DestinationChainSyncToBuffer = 5;
 
         /// <summary>1x10^24 wei = 1 000 000 tokens</summary>
         public BigInteger ReserveBalanceTarget = BigInteger.Parse("1000000000000000000000000");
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -884,29 +917,9 @@ private async Task ProcessConversionMintRequestsAsync()
                 IETHClient clientForDestChain = this.ethClientProvider.GetClientForChain(request.DestinationChain);
 
                 if (!isTransfer)
-                {
-                    this.logger.Info("Processing conversion mint request {0} on {1} chain.", request.RequestId, request.DestinationChain);
```
fassadlr, line 888: I moved this logic into the `PerformReplenishmentAsync` method.

src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -476,17 +505,16 @@ private async Task PollCirrusForBurnsAsync(int blockHeight)
                                 block.GetHash()
                                ));
 
-                            KeyValuePair<string, string> contractMapping = this.interopSettings.GetSettingsByChain(DestinationChain.ETH).WatchedErc20Contracts.First(c => c.Value == receipt.To);
```
fassadlr, line 479: All of this logic only applies if it is inside the `if`.

src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -264,14 +254,40 @@ public async Task InitializeAsync()
...
+                {
+                    this.logger.Warn($"Exception raised when polling for conversion burn & transfer transactions: {e}");
+                }
+            },
+            this.nodeLifetime.ApplicationStopping,
+            repeatEvery: TimeSpans.TenSeconds,
+            startAfter: TimeSpans.Minute);
+
+            // Initialize loop that polls for all transfer (SRC20 to ERC20) requests.
+            this.pollCirrusForTransfersLoop = this.asyncProvider.CreateAndRunAsyncLoop("PollCirrusForBurnsAsync", async (cancellation) =>
```
fassadlr, line 268: Created a separate loop for Cirrus ;)

zeptin (approved):

PR #975, opened 2022-05-10
NFTIndexer: ability to unwatch contracts and validation of contract addresses to add to watched

noescape00:


zeptin (approved):

zeptin:
One thought is that the state repository root can be be interrogated to determine whether a given address is a contract address at all (see `P2PKHNotContractRule` for instance). But I suppose that would preclude pre-emptively adding an address to the indexer prior to the contract being created on-chain; not sure whether that is a significant disadvantage.

PR #986, opened 2022-05-23
[InterFlux] Update local conversion requests state to processed

fassadlr:
We had to remove the "auto-expiring" mechanism for nodes where the conversion request was expired if older than max-reorg blocks. 

Should a node be down it is possible that that node will receive the request and expire it without it actually being processed. Those conversion requests should remain unprocessed until at least one node has set its convesion request state to processed.

The idea here is to run a periodic task that braodcasts a payload asking the nodes for the state they have on a particular conversion request. If it is processed, update our state of that request to processed.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Interop/InteropMonitor.cs
+++ b/src/Stratis.Bitcoin.Features.Interop/InteropMonitor.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using NBitcoin;
 using NLog;
 using Stratis.Bitcoin.AsyncWork;
 using Stratis.Bitcoin.Features.Interop.Payloads;
@@ -15,8 +16,14 @@ namespace Stratis.Bitcoin.Features.Interop
 {
     public sealed class InteropMonitor : IDisposable
     {
+        /// <summary>
+        /// Only attempt to process conversion requests which arrived this amount of block ago or less.
+        /// </summary>
+        private const int ProcessingThresholdBlocks = 2000;
+
         private IAsyncLoop periodicCheckForProcessedRequests;
         private readonly IAsyncProvider asyncProvider;
+        private readonly ChainIndexer chainIndexer;
         private readonly IConversionRequestRepository conversionRequestRepository;
         private readonly IFederationManager federationManager;
         private readonly IFederatedPegBroadcaster federatedPegBroadcaster;
@@ -26,6 +33,7 @@ namespace Stratis.Bitcoin.Features.Interop
 
         public InteropMonitor(
             IAsyncProvider asyncProvider,
+            ChainIndexer chainIndexer,
             IConversionRequestRepository conversionRequestRepository,
             IFederationManager federationManager,
             IFederatedPegBroadcaster federatedPegBroadcaster,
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Interop/Payloads/ConversionRequestStatePayload.cs
```diff
@@ -0,0 +1,71 @@
+﻿using NBitcoin;
+using Stratis.Bitcoin.P2P.Protocol.Payloads;
+using Stratis.Features.FederatedPeg.Conversion;
+
+namespace Stratis.Bitcoin.Features.Interop.Payloads
+{
+    [Payload("conversionrequeststate")]
```
zeptin, line 7: I think these names are limited to a much shorter string?

src/Stratis.Bitcoin.Features.Interop/Payloads/ConversionRequestStatePayload.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.isRequesting = isRequesting;
+        }
+
+        /// <inheritdoc/>
+        public override void ReadWriteCore(BitcoinStream stream)
+        {
+            stream.ReadWrite(ref this.requestId);
+            stream.ReadWrite(ref this.requestState);
+            stream.ReadWrite(ref this.isRequesting);
+        }
```
zeptin, line 53: If we don't serialise the signature as well, then does it actually get sent over the network to the peer?

zeptin, line 53: If we don't serialise the signature as well, then does it actually get sent over the network to the peer?

src/Stratis.Bitcoin.Features.Interop/Payloads/ConversionRequestStatePayload.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.Interop/Payloads/ConversionRequestStatePayload.cs
fassadlr: Oops this is wrong, will fix.

src/Stratis.Bitcoin.Features.Interop/InteropMonitor.cs
```diff
@@ -0,0 +1,98 @@
...
+            },
+            this.nodeLifetime.ApplicationStopping,
+            repeatEvery: TimeSpans.Minute,
+            startAfter: TimeSpans.Minute);
+        }
+
+        private async Task BroadcastConversionRequestStatePayloadAsync(ConversionRequest conversionRequest)
+        {
+            string signature = this.federationManager.CurrentFederationKey.SignMessage(conversionRequest.RequestId);
+            await this.federatedPegBroadcaster.BroadcastAsync(ConversionRequestStatePayload.Request(conversionRequest.RequestId, signature)).ConfigureAwait(false);
```
zeptin, line 89: Maybe we should have logging that each request is being sent? That's minor though

zeptin, line 89: Maybe we should have logging that each request is being sent? That's minor though

zeptin (approved):

PR #994, opened 2022-06-08
[InterFlux] Reprocess Transfer fees to the MultiSig

fassadlr:
The idea here is that we iterate all the mints and burns and check if there is a corresponding CCTS in the store. If not, reprocess the fee to the multisig,

### DIFF
--- a/src/Stratis.Bitcoin.Features.Interop/Controllers/InteropController.cs
+++ b/src/Stratis.Bitcoin.Features.Interop/Controllers/InteropController.cs
@@ -787,11 +787,11 @@ namespace Stratis.Bitcoin.Features.Interop.Controllers
         [ProducesResponseType((int)HttpStatusCode.OK)]
         [ProducesResponseType((int)HttpStatusCode.BadRequest)]
         [ProducesResponseType((int)HttpStatusCode.InternalServerError)]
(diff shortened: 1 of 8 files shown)


src/Stratis.Bitcoin.Features.Interop/Controllers/InteropController.cs
```diff
@@ -767,6 +774,62 @@ public IActionResult DecodeInterFluxTransaction(string hex)
...
+        [Route("reprocessfee")]
+        [HttpPost]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.InternalServerError)]
+        public async Task<IActionResult> ReprocessMultisigFeeAsync([FromBody] ReprocessMultisigFeeModel model)
+        {
+            try
+            {
+                ReprocessFeeResult result = await this.multiSigFeeService.ReprocessFeeAsync(model.RequestId);
```
zeptin, line 794:
```suggestion
                ReprocessFeeResult result = await this.multiSigFeeService.ReprocessFeeAsync(model.RequestId).ConfigureAwait(false);
```

zeptin (approved):

zeptin (approved):

PR #1006, opened 2022-06-23
Add return type specification for some methods in Unity API

madrazzl3:
@noescape00  check plz m8te
probably I'm targeting wrong branch


noescape00:
Retarget PR to release/1.4.0.0

PR #1026, opened 2022-07-22
.NET6 Unit test fixes

fassadlr:


src/Stratis.SmartContracts.CLR.Validation.Tests/ValidationTests.cs
```diff
@@ -551,7 +552,9 @@ public TypeDefinition CompileToTypeDef(string source)
         {
             var mscorlib = MetadataReference.CreateFromFile(typeof(object).Assembly.Location);
             var smartContracts = MetadataReference.CreateFromFile(typeof(Address).Assembly.Location);
-            var runtime = MetadataReference.CreateFromFile(typeof(System.Runtime.AssemblyTargetedPatchBandAttribute).Assembly.Location);
+
+            var assembly = Assembly.Load("System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a");
```
YakupIpek, line 556: Btw it works fine if i load by `Assembly.Load("System.Runtime")` as well and i am not sure which one is more ideal solution.

SatyaKarki (approved):
Seems okay to me.

src/Stratis.SmartContracts.CLR.Validation.Tests/ValidationTests.cs
SatyaKarki: I think, it will be better if we find some other way to get this dynamically from package manager. Otherwise when we upgrade the assembly we need to change this line of code manually as well.

YakupIpek: I think you can't reference that assembly statically so it should loaded at runtime

fassadlr: Also, remember, this is just a test case ;)

fassadlr: But I will change this to Yakup's suggestion

PR #1037, opened 2022-08-10
NFT indexer improvements

noescape00:


zeptin (approved):
It seems to me like this could be optimised quite a lot; if the `ReceiptSearcher` was extended to accept a list of multiple contract addresses, then the blocks would only have to be scanned once each rather than once per contract, as is currently being done

PR #1039, opened 2022-08-15
[API] Add gettransactionsigners endpoint

zeptin:
Adds rudimentary support for determining which pubkey(s) correspond to the private key(s) that signed a given transaction input.

Currently it supports P2PKH and federation multisig transactions. Support for others can be added as needed; it is mostly the scriptPubKey comparisons that need to be amended. We have to take this approach because we don't know what the correct `recId` value is for the pubkey recovery without something to check against (in this case the prevOut's scriptPubKey).

This may also need better handling of the script verify flag values to truly support every transaction type (e.g. cold staking).


zeptin:
Ah, this is targeting the wrong branch, I'll fix that

PR #1052, opened 2022-09-10
Batch receipt searches for NFT indexer

zeptin:
See #1037. 

It may make more sense in future to watch all NFT contracts, rather than to keep having to reindex every time a new one is added.

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts/Wallet/SmartContractTransactionService.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts/Wallet/SmartContractTransactionService.cs
@@ -415,18 +415,21 @@ namespace Stratis.Bitcoin.Features.SmartContracts.Wallet
         {
             var filteredContractAddresses = new HashSet<string>();
 
-            foreach (string contractAddress in contractAddresses)
+            if (contractAddresses != null)
             {
-                uint160 address = contractAddress.ToUint160(this.network);
-
-                byte[] contractCode = this.stateRoot.GetCode(address);
-
-                if (contractCode == null || !contractCode.Any())
+                foreach (string contractAddress in contractAddresses)
                 {
-                    continue;
-                }
+                    uint160 address = contractAddress.ToUint160(this.network);
 
-                filteredContractAddresses.Add(contractAddress);
+                    byte[] contractCode = this.stateRoot.GetCode(address);
+
+                    if (contractCode == null || !contractCode.Any())
+                    {
+                        continue;
+                    }
+
+                    filteredContractAddresses.Add(contractAddress);
+                }
             }
 
             IEnumerable<byte[]> topicsBytes = topics != null ? topics.Where(topic => topic != null).Select(t => t.HexToByteArray()) : new List<byte[]>();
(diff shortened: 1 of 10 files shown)


src/Stratis.Bitcoin.Features.SmartContracts/Wallet/SmartContractTransactionService.cs
```diff
@@ -415,18 +415,21 @@ public List<ReceiptResponse> ReceiptSearch(List<string> contractAddresses, strin
         {
             var filteredContractAddresses = new HashSet<string>();
 
-            foreach (string contractAddress in contractAddresses)
+            if (contractAddresses != null)
```
fassadlr, line 418: Can this list be null?

zeptin: If it is passed as null or empty, then the assumption is that all contracts should be searched for matching logs

src/Stratis.Features.Unity3dApi/LocalCallContract.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        public LocalExecutionResponse LocalCallSmartContract(LocalCallContractRequest request)
+        {
+            ContractTxData txData = this.smartContractTransactionService.BuildLocalCallTxData(request);
+
+            var height = request.BlockHeight ?? (ulong)this.chainIndexer.Height;
+
+            ILocalExecutionResult result = this.localExecutor.Execute(
+                height,
+                request.Sender?.ToUint160(this.network) ?? new uint160(),
```
fassadlr, line 48: Is `Sender` not required?

zeptin: For a local call it seems to be almost unimportant what the actual sender address is. This class is actually copied from one of @quantumagi's pull requests, so any changes would also need to be reflected there too

src/Stratis.Features.Unity3dApi/NFTTransferIndexer.cs
```diff
@@ -188,14 +185,19 @@ public void ReindexAllContracts()
...
                 contractModel.OwnedIDsByAddress = new Dictionary<string, HashSet<long>>();
-                contractModel.LastUpdatedBlock = watchFromHeight;
 
-                this.NFTContractCollection.Upsert(contractModel);
+                updated.Add(contractModel);
             }
 
+            this.NFTContractCollection.Upsert(updated);
+
+            this.UpdateLastUpdatedBlock(0);
```
fassadlr, line 199: I think that 0 should be the `watchFromHeight` ? (i.e. so that it starts at bliock 3 000 000 I think)?

zeptin: Good point, I will amend that

src/Stratis.Features.Unity3dApi/NFTTransferIndexer.cs
```diff
@@ -215,69 +217,74 @@ private async Task IndexNFTsContinuouslyAsync()
...
+                            if (!knownContracts.Contains(receiptRes.To))
+                            {
+                                if (!this.nftContractLocalClient.SupportsInterface((ulong)chainTip.Height, receiptRes.To, TokenInterface.INonFungibleToken))
+                                {
+                                    this.logger.LogTrace("Found TransferLog for non-NFT contract: " + receiptRes.To);
+
+                                    break;
+                                }
+
+                                this.logger.LogTrace("Found new NFT contract: " + receiptRes.To);
```
fassadlr, line 277: Can we `Debug` this log?

fassadlr (approved):

fassadlr (approved):

PR #1059, opened 2022-09-20
Adjust cirrus poll timeout and fix try catch

fassadlr:
Also add checkpoints

### DIFF
--- a/src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
+++ b/src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
@@ -288,7 +288,7 @@ namespace Stratis.Bitcoin.Features.Interop
                 {
                     // In the event that the last polled block was set back a considerable distance from the tip, we need to first catch up faster.
                     // If we are already in the acceptable range, the usual logic will apply.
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin.Features.Interop/InteropPoller.cs
```diff
@@ -284,12 +284,13 @@ public async Task InitializeAsync()
...
 
-                // In the event that the last polled block was set back a considerable distance from the tip, we need to first catch up faster.
-                // If we are already in the acceptable range, the usual logic will apply.
-                await EnsureLastPolledBlockIsSyncedWithCirrusChainAsync().ConfigureAwait(false);
-
                 try
                 {
+                    // In the event that the last polled block was set back a considerable distance from the tip, we need to first catch up faster.
+                    // If we are already in the acceptable range, the usual logic will apply.
+                    // This needs to be in the try catch so that any exceptions doesnt break teh async cloop.
```
zeptin, line 291:
```suggestion
                    // This needs to be in the try catch so that any exceptions don't break the async loop.
```

zeptin (approved):
Verified checkpoints

PR #1060, opened 2022-09-20
Add a new peer connection signalr event and send list of connections in signalr message

SatyaKarki:


src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -89,6 +91,10 @@ public IReadOnlyNetworkPeerCollection ConnectedPeers
 
         private readonly PayloadProvider payloadProvider;
 
+        private readonly ISignals signals;
+
+        public IEnumerable<PeerConnectionModel> peerConnectionModels { get; set; }
```
fassadlr, line 96: You don't need to store a class variable for this. Just create the list in the method an dpublish it 👍

fassadlr (changes requested):
Looks good Satya! Just make that one change :)

fassadlr (approved):

PR #1083, opened 2022-10-31
Publish federation wallet balance via Signalr

fassadlr:


SatyaKarki (commented):
We need ConsensusManagerStatusClientEvent. I saw you have added in new PR. All seems ok.

PR #1084, opened 2022-10-31
[RPC] Add dumpprivkey RPC command

zeptin:


src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletRPCControllerTest.cs
```diff
@@ -29,6 +29,67 @@ public WalletRPCControllerTest()
...
+        {
+            using (NodeBuilder builder = NodeBuilder.Create(this))
+            {
+                CoreNode node = builder.CreateStratisPosNode(this.network).WithWallet().Start();
+
+                RPCClient rpc = node.CreateRPCClient();
+
+                BitcoinAddress address = await rpc.GetNewAddressAsync();
+
+                string apiPubkey = await $"http://localhost:{node.ApiPort}/api"
```
zeptin, line 69: Note that this goes through an unmodified part of the WalletManager to return the associated pubkey, so by comparing the pubkey of retrieved private keys against this string, we can ensure that the code changes for retrieving the private key are not giving incorrect results.

fassadlr (dismissed):

fassadlr (approved):

PR #1096, opened 2022-11-22
Update BTC checkpoints

zeptin:
Not very high priority, but the errors in the json files were getting annoying


quantumagi:
Looks good. Just seeing some CI issues related to test cases that also require updating:

image

quantumagi (approved):

PR #1100, opened 2022-12-29
[RPC] Add createrawtransaction RPC method

zeptin:
This is a somewhat messy RPC to support as the JSON for specifying outputs does not lend itself to conventional models:

```
[
  {                       (json object)
    "address": amount,    (numeric or string, required) A key-value pair. The key (string) is the bitcoin address, the value (float or string) is the amount in BTC
  },
  {                       (json object)
    "data": "hex",        (string, required) A key-value pair. The key must be "data", the value is hex-encoded data
  },
  ...
]
```

It was also found that the RPC client's request handling did not properly support serialising complex models; this has been amended for this RPC call and may make it easier to support others with complex parameters in the future.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
@@ -445,6 +445,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Models
         /// <summary>
         /// Whether to send the change to a P2WPKH (segwit bech32) addresses, or a regular P2PKH address
         /// </summary>
+        [DefaultValue(false)]
         public bool SegwitChangeAddress { get; set; }
 
         /// <inheritdoc />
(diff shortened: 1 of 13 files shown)


src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -445,6 +445,7 @@ public class BuildTransactionRequest : TxFeeEstimateRequest, IValidatableObject
         /// <summary>
         /// Whether to send the change to a P2WPKH (segwit bech32) addresses, or a regular P2PKH address
         /// </summary>
+        [DefaultValue(false)]
```
zeptin, line 448: This is an unrelated but necessary change that will be cherry-picked to the latest feature branch too

noescape00 (dismissed):
LGTM

noescape00 (approved):

PR #1101, opened 2023-01-16
Optimise StraxMain/StraxTest dockerfiles

zeptin:
This results in the images being considerably smaller, by building the daemon in the first stage and then only copying the compiled binaries to the next stage. Since we no longer need the full SDK in the second stage, the resulting image is only about 320mb.

I attempted to build the images using Alpine as well but that introduced several issues for only a modest further space reduction (final image ~200mb).


zeptin:
Incorporated into `release/1.6.0.0` instead

PR #1105, opened 2023-01-20
Add P2WSH handling to wallet database

zeptin:


quantumagi:
image

Why is the extract called twice? Can't we just yield return `txDestination` as before? Does the "change" have any effect?

quantumagi (changes requested):
The `if`'s can be combined.

quantumagi (approved):

zeptin:
In `release/1.6.0.0` we use `GetDestinationFromScriptPubKey` for all known script types. The P2WPKH/P2WSH differentiation is handled automatically by the `PayToWitTemplate` inside the `ScriptAddressReader`, rather than needing separate templates for the two individually.

PR #1119, opened 2023-02-01
.NET 6 Release Rollup

StratisIain:


quantumagi (dismissed):
Agree with this in principle. Did a few high-level checks and it seems good.

zeptin (dismissed):

src/Stratis.CirrusMinerD/Program.cs
```diff
@@ -175,6 +179,10 @@ private static IFullNode BuildStraxNode(string[] args)
                 .UseWallet()
                 .AddSQLiteWalletRepository()
                 .AddPowPosMining(true)
+                .AddSignalR(options =>
+                {
+                    DaemonConfiguration.ConfigureSignalRForCirrus(options);
```
zeptin: This is in the `BuildStraxNode` method?

quantumagi:
The CI issue related to `EnableAutoKickAsync` will be fixed in a follow-up CI:

image

quantumagi:
@StratisIain , actually, from going the other way and attempting to merge `master` into `release/1.4.0.7` its apparent that the `1.4.0.7` stuff is already the latest (when comparing and looking at the differences in the VS IDE). 

I.e. the expectation is that this PR would result in `1.4.0.7` being copied across as the new `master`. So, if you encountered the same (`1.4.0.7` being the latest in each comparison) we should be good on this merge.

PR #1125, opened 2023-03-02
Limit how often the wallet is saved

quantumagi:
There are scenarios where the wallet saves are occurring almost back-to-back and locks the cross-chain process out during that time. This PR addresses that by ensuring at least a 1 minute gap between such save operations.


src/Stratis.Features.FederatedPeg/Wallet/FederationWalletManager.cs
```diff
@@ -289,7 +297,7 @@ public void Start()
                 // save the wallets file every 5 minutes to help against crashes.
                 this.asyncLoop = this.asyncProvider.CreateAndRunAsyncLoop("wallet persist job", token =>
                 {
-                    this.SaveWallet();
+                    this.SaveWallet(true);
```
zeptin, line 300: Should probably do a forced save in the `Stop()` method too

zeptin (approved):

PR #1135, opened 2023-03-30
Remove unused code from FederatedPeg

quantumagi:


src/Stratis.Features.FederatedPeg/Conversion/ConversionRequestKeyValueStore.cs
```diff
@@ -13,8 +13,6 @@ public interface IConversionRequestKeyValueStore : IKeyValueRepository
     {
         List<ConversionRequest> GetAll();
         List<ConversionRequest> GetAll(ConversionRequestType type, bool onlyUnprocessed);
-
-        void Delete(string key);
```
zeptin, line 17: This method does still seem to be in the implementation of the interface though

quantumagi: Yes, but its already present in `IKeyValueRepository`.

zeptin (approved):

PR #1142, opened 2023-05-01
Use deposit blocktime for conversions

zeptin:
This fixes the scenario whereby a Strax deposit (intended for wSTRAX conversion) has a maturing block timestamp that is newer than the Cirrus chain tip, thereby causing the deposit to be ignored.

In order to avoid this, the blocktime of the block the deposit was made in is used instead of the block the deposit matures in. As these conversion deposits never reach the CCTS, the addition of an extra field in the `Deposit` model should not cause problems there.

In order for the `BlockTime` field to be deserialised correctly (but without modifying the `Deposit` constructor throughout the solution), it has been given a public setter.

The return value of `FindApplicableConversionRequestHeader` is additionally explicitly set to `true` or `false` depending on the outcome of the header search.


quantumagi:
Another approach would be to simply calculate the deposit time from the maturity time subject to the type of conversion. That time could then be fed into a slightly modified `FindApplicableConversionRequestHeader`:

`private bool FindApplicableConversionRequestHeader(uint depositBlockTime, out ChainedHeader chainedHeader)`

(In fact the method could now be named more generically since it simply finds the first block after a given point in time.)

The only problem with the above suggestion is that the calculation may not be as straightforward as there have been legacy changes to maturity times BUT a simpler approach of subtracting deposit height from maturity height should reliably give you the correct answer:

```
uint maturityBlocks = MaturedBlockDepositsModel.BlockInfo.BlockHeight - Deposit.BlockNumber;
uint depositBlockTime = MaturedBlockDepositsModel.BlockInfo.BlockTime - maturityBlocks * CounterChainTargetSpacing
```

quantumagi (approved):
LGTM

PR #1144, opened 2023-05-29
Support CrosschainLog

quantumagi:
Adds ability to `Interflux` to incorporate destination account and network from a `CrosschainLog`. SC usage is as follows:
```
    public void MintWithMetadataForNetwork(Address account, UInt256 amount, string metadata, string destinationAccount, string destinationNetwork)
    {
        MintWithMetadata(account, amount, metadata);

        // Perform a cross-chain transfer.
        if (TransferFrom(account, Interflux, amount))
        {
            Log(new CrosschainLog()
            {
                Address = destinationAccount,
                Network = destinationNetwork
            });
        }
    }
```


zeptin:
This looks good, but maybe the changes from #1145 should be incorporated into it as well, as there is some overlap.

zeptin (approved):

PR #1149, opened 2023-05-31
Rename controller method to correct name

zeptin:
Renames a federation controller method to match the SCMI branch, which is the only place it was being used. There was a slight implementation difference in the SCMI version (returning addresses instead of pubkeys) which has also been ported to 1.4.0.7.


quantumagi:
The CCTS was not designed to allow transfers to be deleted. What checks have been performed to ensure that asymmetric or asynchronous deletions will not run counter to the logic surrounding determinacy etc? Is there any test coverage for this? We don't want to resort to copying data folders around to get everything back in sync when deletions run awry. 

Note that there were recent fixes related to a code modification that led to duplicate withdrawal transaction creation. That stabilized the process to the extent that we should probably never have to resort to hacks like these. I guess the reasoning is that the API will only ever be used by responsible individuals that are aware of the risks?

However, given that this PR's main aim is to bring diverting branches in sync I'm approving it subject to any actions you deem necessary to address the above note, now or at a later stage.

quantumagi (approved):

zeptin:
Seems to be in 1.6.0.0 already

PR #1152, opened 2023-06-28
Testnet network patch

quantumagi:
The branch associated with this PR can be used to run nodes on the `StraxTest` and `CirrusTest` private blockchains.

It will be kept up-to-date with changes to the `feature/1.5.0.0-interflux` branch.


YakupIpek (changes requested):
Verifying ECRecover signature is actually should seperated from url parsing. They are different tasks. Secondly parsing url should not belong this common library because it is fully specific for Invoice contract parameters. The Invoice Contract itself should parse parameters if needed.

PR #1153, opened 2023-06-28
[InterFlux] Allocate conversion fees based on confirmation participation

zeptin:
Adds the additional constraint that a multisig federation node must be actively participating in the InterFlux confirmation process in order to receive a portion of the fees thereof.

This is achieved by populating a set of the node pubkeys that confirmed a given transaction at the time it is marked as processed. This tracker is available to the FederatedPeg feature without having any smart contract interface logic inside it.

When a fee is about to be distributed, a list of the request IDs of `n` (currently 5, this can be changed) previous processed InterFlux deposits is retrieved. A set of the pubkeys that participated in confirming these deposits is constructed, nodes that are not present in this set do not receive a portion of the fee.

WIP: Still needs the pubkey/account mappings to be populated


quantumagi:
Hi @zeptin. Two things:
- Disable to old logic related to collateral gaps at activation.
- Use BIP 9 for activation. This was suggested by @StratisIain.

PR #1157, opened 2023-07-07
Update ECRecover and SSAS to support new contracts

quantumagi:


src/Stratis.SCL/ECRecover.cs
```diff
@@ -37,6 +37,35 @@ public static bool TryGetSigner(byte[] message, byte[] signature, out Address ad
...
+        {
+            address = Address.Zero;
+
+            if (message == null || signature == null)
+                return false;
+
+            // NBitcoin is very throwy
+            try
+            {
+                PubKey pubKey = PubKey.RecoverFromMessage(message, Convert.ToBase64String(signature));
```
zeptin: It is important to note that the hash algorithm used here to hash the message under the hood is SHA-256 instead of Keccak256 (which is used in the `TryGetSigner` method). Depending on how the signature was obtained this may be a significant difference, so it should at least be documented to make it obvious. Perhaps the method name should change.

I presume Keccak256 was chosen for similarity to `ecrecover` in Solidity, whereas message signing in the rest of our full node follows the Bitcoin paradigm of SHA-256. Changing the latter is undesirable and changing the former might make interoperability more difficult, although I'm not sure how many current use cases utilise the SCL other than SSAS.

quantumagi: This method is specifically not using hashing due to the way that SignMessage API signs a message in the SSAS context.

zeptin: I am referring to the hashing performed inside `RecoverFromMessage`

quantumagi: Good point. So perhaps we could rename this to `TryGetSignerSHA256` then? I'm reading between the lines that you may also be suggesting moving this method to the SSAS namespace. Its not proper ECRecover after all. That seems reasonable and am I correct in my assumption?

zeptin: That seems reasonable to me

quantumagi: Done.

zeptin (approved):

PR #1166, opened 2023-09-19
Merge release 1.5.0.0 into feature/1.5.0.0-interflux

quantumagi:


src/Stratis.Benchmark/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
zeptin: Looks like an unintentional inclusion

PR #1169, opened 2023-09-20
Attempt to merge 1.5.0.0 into 1.6.0.0

zeptin:
Not ready for review yet


src/NBitcoin/NBitcoin.csproj
```diff
@@ -7,11 +7,11 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <Version>4.0.1.0</Version>
+    <Version>4.0.0.88</Version>
```
zeptin, line 10: Possibly need to decide which version number should be assigned next

